PR #145, opened 2015-11-20
For plugin Tag : Add tag(s) in Tag

Manu1400:
Le plugin Tag ajoute un intitulé Tag dans GLPI, celui-ci apparaît déjà dans les questions FormCreator (Intitulé -> **Gestion des tags** Etiquettes)

Ce patch ne fait qu'ajoute un meilleur support du plugin Tag.

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -789,6 +789,22 @@ class PluginFormcreatorTargetTicket extends CommonDBTM
       if (!$ticketID = $ticket->add($datas)) {
          return false;
       }
+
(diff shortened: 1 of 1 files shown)


neeftarah:
Please, fix your plugins bug and get a stable release before integrate him to another plugin.

Moreover, please explain what your pull request does. I'm not really sure of what it should do, but I think it do it wrong.

And finally, please, update your code to match with latest version of master branch.

neeftarah:
Tag plugin is now integrated in a better way

PR #201, opened 2016-04-22
Fix error in validation for hidden fields with min/max length requirement

Sahrgon:
Hello,

When submitting a form, if a hidden field has a min/max length requirement, the validation shouldn't check its length.

Example :

Take that form with two fields "chp 1" always visible, and "chp 2" always hidden unless "chp 1" equals a certain value.

2016-04-22-113447_987x241_scrot

When you submit the empty form you get these errors :

2016-04-22-113500_326x129_scrot

The first is correct, but the second shouldn't be here because the field "chp 2" is hidden.
This pull request correct that behaviour.

### DIFF
--- a/inc/field.class.php
+++ b/inc/field.class.php
@@ -103,8 +103,7 @@ abstract class PluginFormcreatorField implements Field
 
    public function isRequired()
    {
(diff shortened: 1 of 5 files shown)


DebianRoxx:
Good job !
Works for me !

PR #278, opened 2016-10-09
enable unit tests

dethegeek:


.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+  - '5.4'
+  - '5.5'
+  - '5.6'
+  - '7.0'
+
+before_script:
+  - git clone https://github.com/glpi-project/glpi -b 9.1 ../glpi && cd ../glpi
+  - mv ../formcreator plugins/formcreator
+  - composer install --no-dev
+  - php tools/cliinstall.php --db=glpi-test --user=travis --tests
```
trasher: There is no `--db` option, nor `--user` (looks like it is named `--as-user`).

I guess this should be added, along with a `--pass` one that may be needed on local tests; and that would make script compliant with GLPi' one.

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+
+script:
+  - cd plugins/formcreator
+  - php tools/cliinstall.php --tests
```
trasher: Why to run cliinstall another time here? If it has been runned before script, it should be useless here.

.travis.yml
dethegeek: The current directory for this call to cliinstall.php is the root directory of GLPi. This call is to setup GLPi itself.

.travis.yml
dethegeek: There is now a cliinstall.php in the plugin. It is designed to install the plugin itself for unit tests and may also be used to isntall the plugin from a CLI. The current directory for this call is the root directory of the plugin. 

That's why there a 2 calls to cliinstall.php.

dethegeek:
If the argument --tests is given to the plugin's cliinstall.php then it will setup the constant GLPI_CONFIG_DIR to the GLPi's settings for unit testings. This is less error prone than providing (again) database parameters.

Have a look to lines 25 to 30 in cliinstall.php for the plugin.

dethegeek:
The HTML outputs are produced by the plugin itself when the install process is triggered from GLPi's plugin management page.

The best way to avoid them for unit t esting would be to catch them with output buffering functions ( ob_start() ).

trasher:
OK, thanks you for the clarifications :)

For the HTML output, I'll open a ticket, that can be fixed later... This is not a good way to go, but GLPi and its plugins are used to do that.

As for the cliiinstall, I've noticed that it will use main GLPi test config; so, it is OK also.

Just make a small cleanup in some headers that does not reference the correct plugin/url/whatever and all will be ok.

Thanks for your patience :)

PR #286, opened 2016-10-11
move user area above the services

btry:
The left menu looks like this with these changes

image

I believe there is still something to do to give a better look and feel before merging

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -924,12 +924,14 @@ a.plugin_formcreator_formTile_title {
 }
 
 #plugin_formcreator_avatar {
-   width: 64px;
-   height: 64px;
    margin: 13px 0;
    display: inline-block;
-   background: url('../pics/user.png') no-repeat;
-   background-size: contain;
+}
+
+#plugin_formcreator_avatar img {
+   width: 64px;
+   height: 64px;
+   border-radius: 5px;
 }
 
 #myname.plugin_formcreator_myname {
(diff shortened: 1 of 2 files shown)


orthagh:
I think placing entities/profiles under user shoud be better

btry:
A try with your request gives this

image

I think this may be better to move the entity / profile selectors after the menu like  this, because these will be mainly used by power ful users (form designers ?), and this will allow a rendering ig the service catalog closer to the end users experience. Example below

image

orthagh:
Ok for entity/profiles selector.

I'm not convinced with the current proposal for others parts.
Maybe we can try to integrate icons and user photo/name on the same line as the logo (with breakings lines for some parts)

orthagh:
For Bookmarks, an idea (an old one, in fact): maybe we could offer a full page (and a dedicated text menu on the same level as issues).

btry:
Here is a draft showing my curent orientation about the user avatar and buttons

image

However, I wonder where to show the user name. Maybe under the avatar, aligned to the right of the blue area.

I removed bookmarks because the 3 buttons needs too much height. I'm considering moving the bookmarks as a main menu item.

orthagh:
Maybe i missed to push it, but previously, i changed the user (awesome-) icon for real avatar as we have it in glpi fs ?

You removed it or i didn't push id ?

btry:
I guess you didn't pushed it : I tried to update ma branche a few minutes before pushing my latest changes, and I was up to date

orthagh:
ok :D

PR #292, opened 2016-10-12
Rewrite form validator relation

btry:
A table in the plugin is a relation, but is managed by hardcoded requests only.

- change name of the table to follow GLPi's naming conventions
- Create a class and use it in the whole plugin
- migrate data from the old table to the new one backup the old table)
- add some unit  tests

TODO Add more unit tests before merge

### DIFF
--- a/tests/0010_Integration/Form_ValidatorTest.php
+++ b/tests/0010_Integration/Form_ValidatorTest.php
@@ -1,0 +1,55 @@
+<?php
+class Form_ValidatorTest extends SuperAdminTestCase {
+
+   public function setUp() {
+      parent::setUp();
+
+      $this->formData = array(
+            'entities_id'           => $_SESSION['glpiactive_entity'],
+            'name'                  => 'a form',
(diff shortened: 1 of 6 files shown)


btry:
Bug fixed, but still does not have unit tests. You may review it and merge if it seems OK and you need it quickly.

tests/0010_Integration/Form_ValidatorTest.php
```diff
@@ -0,0 +1,79 @@
...
+            'validation_required'   => 2
+      );
+
+      $this->formDataForUser = array(
+            'entities_id'           => $_SESSION['glpiactive_entity'],
+            'name'                  => 'a form for user validator',
+            'description'           => 'form description',
+            'content'               => 'a content',
+            'is_active'             => 1,
+            'validation_required'   => 1
```
orthagh: idem, class constant

orthagh:
I'll push a commit for my comments.

PR #300, opened 2016-10-18
Add urgency field type

btry:
The urgency field allow a requester to set the urgency of one or several ticket destinations of a form


btry:
Added a pic for urgency field. 

Unit tests coming soon.

PR #301, opened 2016-10-18
big review for service catalogue

orthagh:


btry:
Icon for "book an asset" is missing. 

Except this little point, OK for me

PR #306, opened 2016-10-19
[WIP]Fix meteo

orthagh:
Work in progress, do not merge for the moment


orthagh:
Ok here it is.
It's a pretty big commit, i know.

Desciption (Not in order):
- Last iteration of design / ergonomy, i think it can be the last this time (maybe some little touch needed).
- slinky menu ordered by name
- i removed some parts which, i think was in doulon with core code (footer for example)
- added a recap for 'to validate' tickets
- fix some doublons when querying issues (added a forcegroupby on id key)
- for lefttab layout to avoid some part shifting (in ticket detail)
- remove Jérémy Moreau from displayed author in plugin list. (Maybe we should consider an authors file who credit him ans others, atm, i think he will not answer to questions)
- fixed height of textareas when richtext disabled (the height of textareas and single text field was similar) 

@btry, I have two other improvement needed by the customer (ticket validation searchoption & a observer special question) but we could see that next monday, i'm tired by the week and the trip atm. 

I think, it will be the last week for service catalogue feature.
I suggest a period with a RC tag (released) to avoid regressions and call for tests.

After that we'll be starting new features in master (i'll talk about monday also)

inc/issue.class.php
```diff
@@ -359,7 +380,96 @@ static function getReopenableStatusArray() {
    }
 
    static function getAllStatusArray($withmetaforsearch=false) {
-      return Ticket::getAllStatusArray($withmetaforsearch);
+      $ticket_status = Ticket::getAllStatusArray($withmetaforsearch);
+      $form_status = array('waiting', 'accepted', 'refused');
+      $form_status = array_combine($form_status, $form_status);
+      $all_status = array_merge($ticket_status, $form_status);
```
btry: There is an issue using array_merge here : numeric keys are reindexed from 0 and this triggers a PHP warning line 464. (displayed with debug mode enabled)

Using the array operator + solves the issue.

btry:
Fix added after the merge

PR #316, opened 2016-10-25
Export/import forms

orthagh:
see #314

PR #327, opened 2016-10-27
fix bad column name; requires uprade of the plugin

btry:
Needs tests to ensure I did not missed any use of the column


orthagh:
I refuse this pr, job alredy done in #316

btry:
:+1:

PR #337, opened 2016-11-15
Show forms to validate

btry:
for users in simplified interface : show forms they may validate in  the service catalog


hook.php
```diff
@@ -74,7 +74,7 @@ function plugin_formcreator_addDefaultJoin($itemtype, $ref_table, &$already_link
    switch ($itemtype) {
       case "PluginFormcreatorIssue" :
          $join = Search::addDefaultJoin("Ticket", "glpi_tickets", $already_link_tables);
-         $join = str_replace('`glpi_tickets`', '`glpi_plugin_formcreator_issues`', $join);
+         $join = str_replace('`glpi_tickets`.`id`', '`glpi_plugin_formcreator_issues`.`original_id`', $join);
```
orthagh: on this part, i think we don't have anymore '`glpi_tickets`.`id`' due to the previous replace

PR #340, opened 2016-11-18
Update tests

btry:
fix possible checks on production log files instead of test log files (see phpunit.xml.dist in GLPi)


btry:
bad base branch

PR #343, opened 2016-11-19
prevent email notifications when uninstalling the plugin

btry:
The plugin sends ticket update notifications when uninstalling. We need to revert to a hard SQL query.


btry:
An idea :

I may temporarly set  $CFG_GLPI['use_mailing] = 0 and reset it to its previous state after the job.

Issue #331, opened 2016-11-02
preselection not usefull for us 


Armin83:
In past the radio buttons were not preselected at the branch they are selected by default the first one. Can we deactivate the preselect ?

btry:
Hi 

Here is a patch proposal : https://github.com/btry/formcreator/tree/331_radio_without_default_value

Armin83:
The patch works for us.

btry:
Rework still not started, then we decided to merge the fix anyway in 2.4.1

btry:
merged in the hotfix branch

PR #344, opened 2016-11-21
allow radio buttons without default value

btry:
override parent's implementation which makes default value mandatory

see #331


btry:
With @orthagh we decided to not merge for release 1.4, because it could probably be done a better way after a redesign of fields system.

(here this is an overload of the parent's method to avoit regression due to a previous fix)

PR #349, opened 2016-11-22
remove direct sql queries

btry:


inc/target.class.php
```diff
@@ -137,14 +137,20 @@ public function prepareInputForAdd($input)
...
-                            );";
-                  $DB->query($query);
+                  $targetTicket_actor = new PluginFormcreatorTargetTicket_Actor();
+                  $targetTicket_actor->add(array(
+                        'plugin_formcreator_targettickets_id'  => $id_targetticket,
+                        'actor_role'                           => 'requester',
+                        'actor_type'                           => 'creator',
+                        'use_notification'                     => '1'
+                  ));
+                  $targetTicket_actor = new PluginFormcreatorTargetTicket_Actor();
```
trasher: I do not think this second one is needed, you should you call again add() method.

btry: I do this by preference : always use a fresh instance when I use add() or getFroDB().

PR #361, opened 2016-11-28
fix bad column name

btry:
also fix the same typo in a variable name


btry:
Please double check !

btry:
thanks !

PR #362, opened 2016-11-28
review method do display and search from ticket Summary

orthagh:
I'm using Search for display summary, now we can't have any differences between the summary and the displayed search.

I also factorized criteria for theses search

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -82,10 +82,16 @@ function plugin_formcreator_addDefaultJoin($itemtype, $ref_table, &$already_link
 }
 
 
-function plugin_formcreator_getCondition($table) {
-   if (Session::haveRight('config', UPDATE)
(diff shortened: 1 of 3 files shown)


btry:
Hi

I got the following warning when displaying forms to validate 

```
2016-11-29 20:33:10 [2@workstation-01]
  *** PHP Notice(8): Undefined index: right
  Backtrace :
  ajax/searchoptionvalue.php:113                     
  ajax/searchoption.php:102                          include()
  ajax/searchrow.php:184                             include()
  inc/search.class.php:1873                          include()
  inc/search.class.php:79                            Search::showGenericSearch()
  plugins/formcreator/front/issue.php:29             Search::show()
```

btry:
It seels the issue is not in Formcreator, but GLPi itself

image


Then I approve this PR

PR #366, opened 2016-11-29
saveform() should not contain frontend code

btry:
avoids unwanted output when running unit tests

### DIFF
--- a/front/form.form.php
+++ b/front/form.form.php
@@ -66,7 +66,9 @@ if ($plugin->isActivated("formcreator")) {
          }
 
          // Save form
-         $form->saveForm();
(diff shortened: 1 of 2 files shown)


front/form.form.php
```diff
@@ -66,7 +66,9 @@
          }
 
          // Save form
-         $form->saveForm();
+         if (!$form->saveForm()) {
+            Html::back();
```
trasher: Maybe display an error message?

btry: Messages are generated in the saveForm() method :)

trasher: Ok then :)

PR #376, opened 2016-12-13
css and icon for print form

orthagh:


btry:
TU failed because of a break in develop branch. This is fixed, then the merge will pass the tests

PR #381, opened 2016-12-19
generate uuid for question condition

btry:
question conditions don't have uuid 

Needs a different implementation for develop branch


orthagh:
my 2 cents:
Isn't be better to use an 'add' function and in a prepareIInputForAdd, filling the uuid ?

btry:
Agreed : but we decided we add fixes for 1.4 release.

This is already done in development branch, and that's why the impletnation needs to be done a (better) different way. I added it in the current work for multi conditions.

orthagh:
Ok, sorry to miss that.

PR #383, opened 2016-12-19
Feature/multiple conditions

btry:
multiple conditions for questions

OR and AND operators have the same priority. Needs to be fixed before merge

Contains also some code refactor 

Fix uuid generation for question conditions


btry:
Still need : 
- to hide the forst logic operatir in conditions list, 
- to add a test for logic operator priority

btry:
create and save a question with conditions enabled but without actually add a condition
edit the question => the condition is still enabled but no condition line is displayed. 

To fix

PR #386, opened 2016-12-20
WIP prevent slashes problems

btry:
see #198 

The fix proposal in #198 is for current 1.4 version, and cannot be applied safely on develop branch


btry:
obsoleted due to merge conflcts. new PR #409

Issue #387, opened 2016-12-20
cloning a form with question conditions is not reliable


btry:
Hi

Create a form with 2 questions. One of them has a condition.

Duplicate the form.

Observe in the DB the table for question_condition. You will see the column show_field of the  duplicated condition, has the same value as the original condition. This means the condition refers to a question in an other form. 

![image](https://cloud.githubusercontent.com/assets/14139801/21355039/c89ce976-c6cc-11e6-9baa-82be626c4ca3.png)

PR #389, opened 2016-12-20
fix question_condition duplication when duplicating a form

btry:
**Needs cherry-pick to master too**

see #387


btry:
**Needs cherry-pick to master too**

btry:
cherry-picked into master

PR #390, opened 2016-12-21
Bugfix/duplicate section or question

btry:
Should be merged after #389 (which includes the 2 reverted commits)

I think some code may be factorized before merge.

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -175,7 +175,7 @@ class PluginFormcreatorQuestion extends CommonDBChild
             echo "<span class='form_control pointer'>";
             echo '<img src="' . $CFG_GLPI['root_doc'] . '/plugins/formcreator/pics/clone.png"
                      title="' . _sx('button', "Duplicate") . '"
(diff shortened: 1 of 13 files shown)


inc/question.class.php
```diff
@@ -866,25 +866,33 @@ public static function uninstall()
...
       }
 
-      if (!$this->getFromDB($input[static::getIndexName()])) {
-         return false;
+      // Form questions conditions
+      $rows = $question_condition->find("`plugin_formcreator_questions_id` IN  ('$oldQuestionId')");
+      foreach($rows as $conditions_id => $row) {
+         unset($row['id'],
+               $row['uuid']);
+         $row['plugin_formcreator_questions_id'] = $newQuestion->getID();
```
orthagh: miss show_field ?

btry: Here is how I designed the show_field update process

For form duplication (done in #389) I update all show_fields because the whole form is duplicated. This ensure show_field is not cross-form.

For section duplication (in this PR) I update show_field only if the referenced question ID is duplicated. If not I keep the original question ID.

For single question duplication (in this PR), show_field always references an other question if the same form, then I think it shold not be changed (by the way show_field always different of the being duplicated question ID)

PR #391, opened 2016-12-21
camelcase for method names

btry:
**to cherry-pick into develop too**


btry:
Cherry-picked into develop

Issue #323, opened 2016-10-27
Clean front files


btry:
At least front/formdisplay.php contains some code to do a SQL query. This should move to a class

PR #393, opened 2016-12-21
Feature/323 clean sql

btry:
see #323 : 

remove SQL in /front/*.php files

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -533,6 +533,44 @@ class PluginFormcreatorQuestion extends CommonDBChild
       return $input;
    }
 
+   public function moveUp() {
+      $order         = $this->fields['order'];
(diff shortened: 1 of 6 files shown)


inc/question.class.php
```diff
@@ -533,6 +533,44 @@ protected function deserializeDefaultValue($input) {
       return $input;
    }
 
+   public function moveUp() {
+      $order         = $this->fields['order'];
+      $sectionId     = $this->fields['plugin_formcreator_sections_id'];
+      $otherItem = new static();
+      $otherItem->getFromDBByQuery("WHERE `plugin_formcreator_sections_id` = '$sectionId'
```
orthagh: CommonDBTM::find usage could be better, no ? (it's really a question)

btry: I have to update the found item, then it appears to me better to retrieve directly the object to update it a few lines later.

PR #405, opened 2017-01-02
remove direct sql

btry:
disable mail notification before adding a followup the regular way

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -1505,20 +1505,26 @@ EOS;
 
       // Attach validation message as first ticket followup if validation is required and
       // if is set in ticket target configuration
-      // /!\ Followup is directly saved to the database to avoid double notification on ticket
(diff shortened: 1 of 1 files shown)


inc/targetticket.class.php
```diff
@@ -1505,20 +1505,26 @@ public function save(PluginFormcreatorForm_Answer $formanswer) {
...
             $message.= "\n".addslashes($formanswer->fields['comment']);
          }
 
-        $query = "INSERT INTO `glpi_ticketfollowups` SET
-                     `tickets_id` = $ticketID,
-                     `date`       = NOW(),
-                     `users_id`   = {$_SESSION['glpiID']},
-                     `content`    = \"$message\"";
-         $DB->query($query);
+        // Disable mail notification
```
orthagh: maybe, we could pick up the removed comment (l1508) to explain why we disable notifications.

PR #406, opened 2017-01-02
Bugfix/401 textarea newline strip

btry:
see #401 

To be cherry-picked into develop

### DIFF
--- a/inc/fields/textarea-field.class.php
+++ b/inc/fields/textarea-field.class.php
@@ -21,7 +21,7 @@ class textareaField extends PluginFormcreatorField
          if ($CFG_GLPI["use_rich_text"]) {
             echo plugin_formcreator_decode($this->getAnswer());
          } else {
-            echo nl2br($this->getAnswer());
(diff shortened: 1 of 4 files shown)


setup.php
```diff
@@ -177,17 +177,19 @@ function plugin_init_formcreator ()
  * @param  String    $string  The string to encode
  * @return String             The encoded string
```
orthagh: we miss a docblock comment

inc/fields/textarea-field.class.php
```diff
@@ -21,7 +21,7 @@ public function displayField($canEdit = true)
          if ($CFG_GLPI["use_rich_text"]) {
             echo plugin_formcreator_decode($this->getAnswer());
          } else {
-            echo nl2br($this->getAnswer());
+            echo str_replace(array("\\r\\n", "\\r", "\\n"), "<br />", $this->getAnswer());
```
orthagh: i'm not very pleased by this change.
If possible, we should store the data without double antislash.
nl2br is a good function, imo

btry:
An other proposal here #410

orthagh:
replaced by #410

Issue #384, opened 2016-12-20
First \ is not displayed in questions


satyan01:
Hello

i would like to add a link to a file in a question but the first \ disappears when I save the question

In fact all \ are deleted after saving

![image](https://cloud.githubusercontent.com/assets/15144026/21345222/a63dfc26-c69f-11e6-9c0a-cab301aa20fe.png)

btry:
hi

I think this is a duplicate of #198 (in french).

is your link a description of a question ?

satyan01:
It is in description area

PR #409, opened 2017-01-03
fix slashes problems

btry:
see #198 
This is an update of https://github.com/btry/formcreator/tree/198_fix_slashes
to avoid merge clnflicts


btry:
Should fix #384 too

PR #477, opened 2017-02-06
add category question

btry:
with background code refactor on urgency


orthagh:
Some comments:
- Add an option to set the category manually without template and question. We aim, i think, to drop template option or minimise it to the minimum (set to final ticket for hidden fields)
- As we have now PluginFormcreatorTargetChange, maybe we should think about a Target Super class with common code (in another pr)

orthagh:
And fix travis build :)

btry:
Added ability to set a specific value for category. For a better consistency, I added the same feature for urgency.

orthagh:
Seems good. Last comment and it will be ok i think.
Could you ordering the fields like this:
- Entity
- Ticket template | Due Date
- Category
- Urgency

btry:
You missed the Tags section. Where do you want to see it in your list ?

orthagh:
last one, didn't saw it because the plugin isn't enabled in my side.

PR #479, opened 2017-02-07
WIP - drop header feature

btry:
with the new design of 2.4.0, the header is no longer used, but the table is still in the schema


btry:
These changes will take place in #480. 

Cleanup of the table already added glpi_plugin_formcreator_headers (its class too)

btry:
PR updated,

see hook.php

PR #480, opened 2017-02-07
Renew install / upgrade system

btry:
solve #453 

This PR assumes #479 was merged.

This PR needs update after #477 validation and merge

PR #434, opened 2017-01-16
use coding standard package

btry:
use GLPI coding standard

PR #488, opened 2017-02-09
Feature/external tools

trasher:


btry:
It invalidates #434 (because it uses the new tools package). 

The first commit of this PR is obviously OK, but the 2 last fixing CS may cause conflicts with the #480 .

trasher:
The second commit is just an automated run of phpbf from the `robo code:cs` call; this is quite simple to cancel an run again.
The last one is manual changes, but there are only a few ones, even if it conflits something, it should not be a huge one.

btry:
@trasher : did you intentionnaly add composer.lock in the PR ?

btry:
#519 merged ! finally ! (and invalidates this one)

PR #502, opened 2017-02-15
Form export

Alibaskins:
Hello,

It would be fine to be able to export form in pdf (for review for exemple).
Or maybe there is already this function ?

Regards,
Ali


btry:
Hi 

Please open an issue instead of a PR for you feature requests or bug reports. The feature you requested will be available in 2.5.0

Issue #522, opened 2017-03-07
Error during migration


yllen:
From 084.2.1.5 to 2.4.0

*** MySQL query error:
  SQL: ALTER TABLE `$table` ADD `name` VARCHAR(255) NOT NULL AFTER `id`;
  Error: Table 'glpi91.$table' doesn't exist
  Backtrace :
  plugins/formcreator/inc/form_answer.class.php:885

btry:
Fixed, thank you

PR #523, opened 2017-03-07
Fix migration - #522

yllen:


btry:
Thank you for spotting this. 

Fixes for version 2.4.0 should go in the branch **htofix/2.4.1**.

I committed d5881d95af3dd8a62de68e316c5bffaa9116094b to do an equivalent fix.

PR #535, opened 2017-03-15
fix not displayed ticket in service catalog

btry:
When a user answers a form and the generated ticket has an ID different to the forms_answers ID then the user cannot show the ticket in the service catalog.

This fix was pushed in hotfix/2.4.1 branch for testing purpose. Needs confirmation that this fix the issue before merging into develop.


btry:
fix reported successful.

PR #539, opened 2017-03-17
Message ack for public

orthagh:


front/formdisplay.php
```diff
@@ -3,9 +3,14 @@
 
 // Check if plugin is activated...
 $plugin = new Plugin();
+$form   = new PluginFormcreatorForm();
+
+PluginFormcreatorForm::header();
+
+if($plugin->isActivated("formcreator")
```
btry: I think we should detach the check for the status of the plugin and move it before affectation of $form. If not, a fatal error may occur

PR #553, opened 2017-03-24
fix #552

### LINKED ISSUE #552: Error message appearing on non debug-mode in english
Bonjour,
Je suis actuellement en train d'essayer tous les types de question possible et j'ai remarqué un bug mineur / anomalie.
Dans un formulaire si l'on ajoute une question de type "intitulé" de base le champ "Valeur par défaut" est vide, on ne peut pas le remplir. Si l'on choisis un intutilé, par exemple un lieu, le champ "Valeur par défaut" apparait bien et on peut le remplir. Néanmoins si l'on selectionne "rien", c'est à dire "---" ,dans le champ type d'intitulé (ici cela remplacera "Lieux") un message d'erreur en anglais s'affiche :
2017-03-24 11_43_11-glpi - formulaires - 1

Ce message s'affiche même si l'on est pas en mode debug et si notre langue choisis est autre que l'anglais.
_PS: Je sais que c'est un problème mineur, je ne fais que du feedback sur la verison actuel du plugin pour que la suivante soit niquel !_

btry:


ajax/dropdown_values.php
```diff
@@ -3,13 +3,14 @@
 
 Session::checkRight("entity", UPDATE);
 
-if(class_exists($_REQUEST['dropdown_itemtype'])) {
+if($_REQUEST['dropdown_itemtype'] == '0' || !class_exists($_REQUEST['dropdown_itemtype'])) {
+   $options = array(
+         'display_emptychoice'   => true,
+   );
+   Dropdown::showFromArray("dropdown_default_value", array(), $options);
```
orthagh: We can certainly set third parameter inline, no ?

```php
Dropdown::showFromArray("dropdown_default_value", [], ['display_emptychoice' => true]);
```

btry: Indeed, I checked and this is short enough to not hit ehe line length limit. 

I prever array(...) instead of [...] for now because the plugin is still compatible with old GLPI versions.

PR #562, opened 2017-04-03
cleanup dead code, rename js folder

btry:
scripts/combobox.js.php  is not used anywhere
scripts/forms-validation.js.php has one function always returning true

rename scripts folder into js


scripts/forms-validation.js.php
```diff
@@ -1,18 +0,0 @@
-<?php
-include ('../../../inc/includes.php');
-header('Content-type: text/javascript');
-?>
-
-function validateForm(form) {
```
orthagh: are you sure about validateForm removal? 

cf: https://github.com/pluginsGLPI/formcreator/blob/master/inc/form.class.php#L838

scripts/combobox.js.php
```diff
@@ -1,136 +0,0 @@
-<?php
```
orthagh: for this part, we have some leftovers in css/style.css

scripts/forms-validation.js.php
btry: You're right, I cancel removal of this file

btry: yes : the content of the function is commented out to always return true

scripts/combobox.js.php
btry: I removed a few CSS based on the removed code, which I don't find anywhere else in the plugin.

Issue #548, opened 2017-03-22
Activate Formular error message during load page


CCoruble:
Hi, 
I added a question to a formular and wanted to "activate" it by selecting him, go on "Action" "Modify" "Activate".
When clicking "send" the page did reload and a really quick message during the loading of the page appeared :
![2017-03-22 08_59_02-glpi - modification massive](https://cloud.githubusercontent.com/assets/24699373/24187699/76c3b166-0ede-11e7-9f6c-c2a8a0f6e2c3.png)
After this the formular wasn't set to "active", the activation failed.

But the activation succeded I used the parameter "Active" and set it to "Yes" :
![2017-03-22 09_09_50-glpi - forms - 5](https://cloud.githubusercontent.com/assets/24699373/24187899/6baa2912-0edf-11e7-9f46-eaf8b59fcc95.png)

btry:
Hi

I found a few weeks ago the massive action is not correct. To enable or disable a form the values should be **Yes ** or **No**. 

Currently it show the following
![image](https://cloud.githubusercontent.com/assets/14139801/24188112/568111d0-0ee0-11e7-9ebc-fa8f2193b30f.png)

I planned to review and fix massive actions, as soom as more critical issues are fixed.

btry:
duplicates #441

btry:
Hi

You may try this patch. It should repair the massive action problem you reported 
https://github.com/pluginsGLPI/formcreator/pull/563

PR #563, opened 2017-04-03
Feature/441 massive action problems

btry:
Bad label for inactive / active dropdown
massive actions  not working see #548 (why **'criteria[0][value]'** for a HTML field name ?)


btry:
this PR addressing a bug, I rebased it to 2.4.2

PR #572, opened 2017-04-05
Revert "forcegroupby on name in issues"

btry:
Reverts pluginsGLPI/formcreator#571


btry:
new fix based on the bad one. This revert becomes useless

Issue #433, opened 2017-01-16
Add general config / profile config


pVesian:
Hi,

It would be nice to be able to deactivate the link added in the main menu bar, either in a general config or on a profile-based level.

Thanks!

btry:
Hi

Which link you're talking about ?

pVesian:
Sorry, I messed up my previous sentence a bit, it's edited now. I'm talking about the link that appears in the main menu bar ("Formulaires" in FR).

btry:
I think this is the breadcrumb here ? 

![image](https://cloud.githubusercontent.com/assets/14139801/22017604/4a83e446-dcab-11e6-8037-5692159501e1.png)

pVesian:
No, it's the "Forms" link in the menu, not the breadcrumb. It appears for any profile (hotliner for example). The link for this link is to plugins/formcreator/front/formlist.php.
![forms_menu_link](https://cloud.githubusercontent.com/assets/3509962/22021284/93c40c2e-dcbc-11e6-91a6-b52302c4ff82.png)

btry:
Is it displayed for profiles who don't have access to any form ?

pVesian:
Yes it is, even when there is no form at all.

btry:
I think it makes sense to hide the link when no forms are available.

Availability of forms is manageable on a per profile basis. If you don't already know this : edit a form, open the tab **Targets** set **Restricted access*, save the change, tick the profiles able to access the form, and save again. 

Then I suggest the plugin should simply count the number of available forms. If there is at least one form, the link is displayed, otherwise, it is hidden.

pVesian:
It would indeed be a good solution. The count would have to be done on the list of available forms for the current user.

davidiasass:
Help me please?

The form does not appear in the menu item.
Appears only in the user profile

dethegeek:
Hi @davidiasass

Please open an other issue to discuss about your problem

pVesian:
Hi @btry, is there any news on this? Thanks.

btry:
Hi

I added the feature in version 2.5.0. 

You may want to try it on a testing environment and feedback before we release this version

btry:
Hi

The feature is added. Having no feedback, I close this ticket.

PR #598, opened 2017-05-09
hide forms menu if no form available

btry:
In the helpdesk, hide the forms menu if no form available

1 form available
image

No form available
image

### DIFF
--- a/js/scripts.js.php
+++ b/js/scripts.js.php
@@ -54,12 +54,12 @@ jQuery(document).ready(function($) {
    });
 
    <?php
-      if (isset($_SESSION['glpiactiveprofile']['interface'])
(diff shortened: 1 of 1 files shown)


js/scripts.js.php
```diff
@@ -55,8 +55,10 @@
 
    <?php
    if (isset($_SESSION['glpiactiveprofile']['interface'])
-            && ($_SESSION['glpiactiveprofile']['interface'] == 'helpdesk')) {
-      echo "$('#c_menu #menu1:first-child').after(link);";
+       && ($_SESSION['glpiactiveprofile']['interface'] == 'helpdesk')) {
```
orthagh: instead of double if, maybe you can add a ```&&```

btry:
see #433

PR #622, opened 2017-06-13
check php version

btry:
Some users may run GLPI with too old PHP version. Check PHP version in the plugin also.


btry:
see #620

orthagh:
ok for the change globally, check travis error btw

PR #642, opened 2017-06-21
fix typo and rename this var

btry:
$options should be $soptions; but $soptions is a rather poor var name


btry:
#643 rebases this fix onto develop (better than GH)

PR #646, opened 2017-06-26
fix blank print page

btry:
Fix blank print page


btry:
see #621

btry:
You're right, with others layouts it does not works. Fixing...

PR #652, opened 2017-06-26
To avoid PHP notice

yllen:
http://forum.glpi-project.org/viewtopic.php?id=158109

### DIFF
--- a/inc/form_answer.class.php
+++ b/inc/form_answer.class.php
@@ -245,7 +245,7 @@ class PluginFormcreatorForm_Answer extends CommonDBChild
 
       // prepare params for search
       $item          = new PluginFormcreatorForm_Answer();
-      $soptions      = $item->getSearchOptions();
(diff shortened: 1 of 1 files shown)


btry:
hi

Thank you, but the fix is already done in develop. 

Tis was a typo, and to avoid it I renamed the var as $soptions is error prone for my eyes

yllen:
Are you sure?
For doing this pull request i open the file directly in code section of this github.

btry:
Yes, I'm sure

see here 
https://github.com/pluginsGLPI/formcreator/blob/develop/inc/form_answer.class.php#L246

this fix will be available in the next release.

yllen:
OK i understand.
Your new version is not in master branch but in develop branch.
What is master branch?

btry:
master branch is the latest release. When a new version is released, master branch is updated, and only in this case. See git flow branching model

PR #654, opened 2017-06-27
GLPI 9.2 - Feature/fix js glpi 9.2

btry:
With GLPI 9.2 scripts.js.php is loaded too late and calls to fucntions it contains occur before they're are available.

This produces many JS errors when displaying a form to fill. (form preview not impacted by  the bug)


btry:
Validated by @flegastelois

Issue #435, opened 2017-01-17
Remove eval() in fields.class.php


btry:
there is eval'd code and we can avoid this; no matter this increases the source size.

btry:
done.

PR #666, opened 2017-07-17
2.6 refactor question conditions

btry:
To get rid of debug pain on questions conditions.

This also is the beginining of a redesign of fields impleemtnation

Solves #435


inc/fields.class.php
```diff
@@ -115,14 +116,12 @@ public static function showField($field, $datas = null, $edit = true) {
     * @return  boolean                 Should be shown or not
     */
    public static function isVisible($id, $values) {
-      global $DB;
```
btry: unused global

inc/fields.class.php
```diff
@@ -179,61 +178,42 @@ public static function isVisible($id, $values) {
             return false;
          }
 
+         // TODO: find the best behavior if the question does not exists
```
btry: use a factory to instanciate and delegate to them the comparison job

inc/fields/actorfield.class.php
```diff
@@ -70,6 +71,10 @@ public function displayField($canEdit = true) {
                      }
                   })
                   $("#actor_formcreator_field_' . $this->fields['id'] . '").select2("readonly", ' . $readonly . ');
+                  $("#actor_formcreator_field_' . $this->fields['id'] . $rand . '").on("change", function(e) {
```
btry: bugfix: the actors were not sent to the server when evaluating visibility of fields

inc/fields/actorfield.class.php
```diff
@@ -27,17 +28,17 @@ public function displayField($canEdit = true) {
       // Value needs to be non empty to allow execition of select2's initSelection
       echo '<input
                type="hidden"
-               name="formcreator_field_' . $this->fields['id'] . '"
-               id="actor_formcreator_field_' . $this->fields['id'] . '"
-               value=" " />';
```
btry: value=" " was source of problems when Select2 component populates this atribute. IDs of users were prefixed and/or suffixed by a comma leading this kind of result:

,2,
,2,5
2,5,

inc/fields/actorfield.class.php
```diff
@@ -27,17 +28,17 @@ public function displayField($canEdit = true) {
...
+               name="formcreator_field_' . $this->fields['id'] . $rand . '"
```
btry: bugfix: the HTML id of the field was not protected by a $rand

btry:
evaluationg lower than is dangeroud because greaterthan may return false in some specific case

https://github.com/pluginsGLPI/formcreator/pull/666/files#diff-a49544257aa68ea39fcfd83e4ed0adacR219

is it useful to evaluate a scalara against a full list ?
https://github.com/pluginsGLPI/formcreator/pull/666/files#diff-a49544257aa68ea39fcfd83e4ed0adacR231

btry:
This should be better with this way to throw and catch exception. 

An other idea may be to forbid inequalities on some not relevant field types. For examle comapring if a user is greater or less than an other is meaningless. I implemented more because I don't know if someone somewhere do this already.

btry:
obsoleted

PR #668, opened 2017-07-18
fix show / hide select field

btry:
Fix HTML entities problem with select fields which should show or hide on condfition when a form is displayed


btry:
Fix validated by the reporter of the issue.

orthagh:
ok for the reporter.
This is particulary hard to review because i'm can't be sure these sort of fixes could not break some other parts, you see.

btry:
I agree that's why I want to remove all these string escaping mess and rewrite the evaluation system nicely numbered #666

orthagh:
:D good timing

PR #686, opened 2017-08-01
limit displayed columns on form answers tab of a form

btry:


btry:
validated by the reporter

PR #693, opened 2017-08-07
2.6.0 Feature/refactor field edition

btry:
refactor creation / edition of fields

Will help to use JSON instead of \r\n or semicolon separated data

### DIFF
--- a/inc/fields/ldapselectfield.class.php
+++ b/inc/fields/ldapselectfield.class.php
@@ -15,15 +15,7 @@ class PluginFormcreatorLdapselectField extends PluginFormcreatorSelectField
             return array();
          }
 
-         if (!function_exists('warning_handler')) {
-            function warning_handler($errno, $errstr, $errfile, $errline, array $errcontext) {
-               if (0 === error_reporting()) {
-                  return false;
-               }
-               throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
-            }
-         }
-         set_error_handler("warning_handler", E_WARNING);
+         set_error_handler('plugin_formcreator_ldap_warning_handler', E_WARNING);
 
          try {
             $tab_values = array();
@@ -86,16 +78,7 @@ class PluginFormcreatorLdapselectField extends PluginFormcreatorSelectField
             $attribute     = array();
          }
 
-         // Set specific error handler to catch LDAP errors
-         if (!function_exists('warning_handler')) {
-            function warning_handler($errno, $errstr, $errfile, $errline, array $errcontext) {
-               if (0 === error_reporting()) {
-                  return false;
-               }
-               throw new ErrorException($errstr, 0, $errno, $errfile, $errline);
-            }
-         }
-         set_error_handler("warning_handler", E_WARNING);
+         set_error_handler('plugin_formcreator_ldap_warning_handler', E_WARNING);
 
          try {
             $ds            = $config_ldap->connect();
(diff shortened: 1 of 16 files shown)


inc/fields/textareafield.class.php
```diff
@@ -1,5 +1,5 @@
 <?php
-class PluginFormcreatorTextareaField extends PluginFormcreatorField
+class PluginFormcreatorTextareaField extends PluginFormcreatorTextField
```
btry: To deduplicate code between text and  textarea fields

inc/fields/textareafield.class.php
```diff
@@ -25,50 +25,10 @@ public function displayField($canEdit = true) {
       }
    }
 
-   public function isValid($value) {
```
btry: duplicated method

inc/fields/textareafield.class.php
```diff
@@ -25,50 +25,10 @@ public function displayField($canEdit = true) {
...
-
-      // All is OK
-      return true;
-   }
-
    public static function getName() {
       return __('Textarea', 'formcreator');
    }
 
-   public static function getPrefs() {
```
btry: duplicated method

inc/fields/ldapselectfield.class.php
```diff
@@ -70,6 +70,54 @@ public static function getName() {
...
+
+         if (!empty($input['ldap_attribute'])) {
+            $ldap_dropdown = new RuleRightParameter();
+            $ldap_dropdown->getFromDB($input['ldap_attribute']);
+            $attribute     = array($ldap_dropdown->fields['value']);
+         } else {
+            $attribute     = array();
+         }
+
+         // Set specific error handler to catch LDAP errors
```
orthagh: one thought here, if its a "specific handler for ldap errors", maybe you can name it according to that.
Ex:  ``fc_ldap_warning_handler``

inc/question.class.php
```diff
@@ -259,146 +259,25 @@ private function checkBeforeSave($input) {
...
-            }
-         }
-      }
-
-      if ($input['fieldtype'] == 'textarea' || $input['fieldtype'] == 'text') {
-         if (isset($input['default_values'])) {
-            $input['default_values'] = addslashes($input['default_values']);
-         }
-      }
+      $fieldType = 'PluginFormcreator' . ucfirst($input['fieldtype']) . 'Field';
```
orthagh: maybe replace ``$fieldType`` by ``$fieldClass``

orthagh:
Apart the two minor issues above, it's a great work :+1: , the code is now more clearer and you removed a bloated function.

inc/fields/ldapselectfield.class.php
btry: This function appears in the iff only because I moved its code. OK to rename it here

inc/question.class.php
btry: $input['fieldtype'] contains only a subpart of the class name.  Also, the subpart is a type of field in a similar meaning as we can find in <input> HTML tag 

```html
<input type="text" name="lastname">
```

I would keep the current name because I think it is more accurate.

orthagh: i mentionned the class var instancied below.
``$field = new $fieldType``

I completely agree the key ``$input['fieldtype']`` is well named.

btry: Oh, sorry !

I prefer ```$fieldObject```. done

Issue #506, opened 2017-02-17
Regex in additional validation field non-ascii characters


gotanbl:
I need to regex some non-ascii characters like čćžđš (bosnian|serbian|croatian special characters) .
For example:
 /[\x{0161}\x{0160}\x{017d}a-z]+/u
\x{0161} represent š
\x{0160} represent Š
\x{017d} represent Ž
stream above represent [šŠŽa-z]+.
After I put it in text field for additional validation and save it, formcreator saved it as
/[x{0161}x{0160}x{017d}a-z]+/u$/
And this not work!
Error in php_errors.log:
*** PHP Warning(2): preg_match(): Unknown modifier '$'
  Backtrace :
  :                                                  
  .../formcreator/inc/fields/text-field.class.php:21 preg_match()
  plugins/formcreator/inc/form.class.php:1088        textField->isValid()
  plugins/formcreator/front/form.form.php:74         PluginFormcreatorForm->saveForm()

btry:
In the source code, the regex is forced to begin with **/^** and to finish with **$/**

Because  your regex terminates with **u**, **$/ is appended. I think you want to use UTF-8 ( http://php.net/manual/fr/reference.pcre.pattern.modifiers.php ) . Can you find a workaround to not use any modifier ?

gotanbl:
Yes, I want to use UTF-8. 
I am not find any other solution, yet. 
I will try to find, but I am not shure is it possible.

gotanbl:
It is possible with \p{L}+      matches any kind of letter from any language.
But also do not work! No errors in php_errors.log.

btry:
Then I think I'll have to modify (maybe drop ?) the checks on regex to allow modifiers.

btry:
Hi

Please have a look on this PR

https://github.com/pluginsGLPI/formcreator/pull/701/files

This would help you by giving you the fill responsibility (and power) on the regex validation.

btry:
Hi 

The PR i submitted to you is merged. I hope it will satisfy your need. If not, please open a new issue.

PR #701, opened 2017-08-21
give to admins full power of regex

btry:
see #506 : need more power for admins with regex validation.


install/update_2.5_2.6.php
```diff
@@ -0,0 +1,33 @@
...
+   $table = PluginFormcreatorQuestion::getTable();
+   foreach ($rows as $id => $row) {
+      // Avoid notice when validating the regex
+      set_error_handler(function($errno, $errstr, $errfile, $errline, $errcontext) {});
+      $isValid = !(preg_match($row['regex'], null) === false);
+      restore_error_handler();
+
+      if (!$isValid) {
+         // The regex is invalid (likely not updated yet)
+         $updatedRegex = addslashes('/^' . $row['regex'] . '$/');
```
orthagh: according to this part https://github.com/pluginsGLPI/formcreator/pull/701/files#diff-bc7cf4201f52932596eb92080c6e915aL406
it seems the previous behavior was a little bit more complex.
We can have these case:

- no prefix
- prefix ``/``
- prefix ``/^``

same with suffix.

btry: When an admin adds a regex validation,  the previous behavior is to detect if the regex has delimiters and anchors. If one of them was missing, then the plugin added them automatically before saving.

So, thee regexes in DB in versions up to 2.5.x always start with "/^" and ends "$/".

Actually, your comment makes me realize this upgrade code is useless. I need to remove it.

btry: By the way, as the upgrade becomes empty, I can revert the upgrade code I took from an other PR. This will avoid useless merge conflicts

orthagh: ok let me know when i need to recheck

PR #703, opened 2017-08-21
2.6.0 - composite  tickets

btry:
Needs unit tests

image

image


orthagh:
by clicking on save button after addition of a linked ticket, we have this error appearing (no cons apparently)
image

btry:
I cannot reproduce. 

The message occurs because the linked ticket or linked  target ticket you choosed does not exists. In your tests it is unlikely something deletes t he item between the instant you select it and the instant the link is established...

btry:
bug : upload de document dvia un question type fichier => document associé à tous les tickets.

PR #707, opened 2017-08-25
simplify entity restriction and enhance performance

btry:
It seems there is a bad design here, and it heavily impacts performance when GLPI has a lot of entities

This PR should fix it

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -427,12 +427,7 @@ class PluginFormcreatorForm extends CommonDBTM
 
       // If the form is recursive, authorize the validators in sub-entities
       // If it isn't, only the validators of the entity of the form
-      if ($this->isRecursive()) {
-         $entites = getSonsOf('glpi_entities', $this->getEntityID());
-      } else {
-         $entites = $this->getEntityID();
-      }
-      $subentities = getEntitiesRestrictRequest("", 'pu', "", $entites, true, true);
+      $subentities = getEntitiesRestrictRequest("", 'pu', "", $this->getEntityID(), $this->isRecursive(), true);
 
       // Select all users with ticket validation right and the groups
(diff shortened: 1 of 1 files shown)


inc/form.class.php
```diff
@@ -336,12 +336,7 @@ public function showForm($ID, $options=array()) {
 
       // Si le formulaire est récursif, on authorise les validateurs des sous-entités
       // Sinon uniquement les validateurs de l'entité du formulaire
-      if ($this->isRecursive()) {
-         $entites = getSonsOf('glpi_entities', $this->getEntityID());
-      } else {
-         $entites = $this->getEntityID();
-      }
-      $subentities = getEntitiesRestrictRequest("", 'pu', "", $entites, true, true);
+      $subentities = getEntitiesRestrictRequest("", 'pu', "", $this->getEntityID(), $this->isRecursive() == 1, true);
```
orthagh: ok for replacement

But why use ``== 1``, seems superflu

wawax: seems ok to me also. Same comment as @orthagh

inc/form.class.php
```diff
@@ -336,12 +336,7 @@ public function showForm($ID, $options=array()) {
 
       // Si le formulaire est récursif, on authorise les validateurs des sous-entités
       // Sinon uniquement les validateurs de l'entité du formulaire
```
wawax: The comment should be in english

inc/form.class.php
btry: == 1 has two purposes : cast into a boolean, and also ensure true is returned only when the column is_recursive contains 1.
I think ```$this->isRecursive() != 0``` is even better (this is an habit I adopted in Flyve MDM plugin for example). I avoid use of empty() to keep things simple.

orthagh: you do exactly the same thing as php core.
If you want to explicitly cast (i still think it's useless here), do it but with a real cast syntax.

btry:
Planning to add unit tests with atoum, for 2.7.0

btry:
Outdated by code refactors

PR #713, opened 2017-09-05
remove dead code

btry:


hook.php
```diff
@@ -89,9 +89,6 @@ function plugin_formcreator_addDefaultWhere($itemtype) {
    $table = getTableForItemType($itemtype);
    switch ($itemtype) {
       case "PluginFormcreatorIssue" :
-         $condition_fanwser = plugin_formcreator_getCondition("PluginFormcreatorForm_Answer");
-         $condition = "`$table`.`sub_itemtype` = 'PluginFormcreatorForm_Answer'
```
btry: this affectation is overwritten thus not used

PR #723, opened 2017-09-18
fix #705 by allowing the value 0 when GLPI object field is for an entity

### LINKED ISSUE #705: bad validity evaluation for an entity dropdown
Use the following question; note it is mandatory
image

When filling a form and selecting the root entity, the question is detected invalid because it is not filled.

enpty(0) evaluated to  true despite the field is filled.

btry:


inc/fields/glpiselectfield.class.php
```diff
@@ -20,6 +20,20 @@ public function prepareQuestionInputForSave($input) {
       return $input;
    }
 
+   public function isValid($value) {
+      // If the field is required it can't be empty (0 is a valid value for entity)
+      if ($this->isRequired() && empty($value) && ($value == '0' && $this->fields['values'] != Entity::class)) {
```
orthagh: do not use `empty`, it a bad function in php, you should list all case where a select is not selected (0 like after, maybe '---', etc)

orthagh: for information, `(empty($var))` is like doing `($var == false)`

btry: This is a technical debt. I overloaded the method from the parent class. I prefer not to change its way too much to avoid regressions.

I'll probably drop use of empty() when I'll refactor all the fields classes. For now I still focus on bugfixes and finishing pending PRs

orthagh: ok fine, review ok in this case

PR #757, opened 2017-09-30
fix(ui): require JS tinymce when displaying target ticket and rich text is enabled

### LINKED ISSUE #754: Destination: specific entity [9.2]
No additional dropdown is loaded, using GLPI 9.2 and the RC release of FormCreator. 

In the old version you would get a addtional dropdown where you could specify extra information.
I can see it not working for the specific entity but I think there's more.. the plugin is acting odd.

btry:
fix #754

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -165,6 +165,10 @@ function plugin_init_formcreator() {
             $PLUGIN_HOOKS['add_javascript']['formcreator'][] = 'js/scripts.js.php';
          }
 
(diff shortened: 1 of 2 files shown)


setup.php
```diff
@@ -165,6 +165,10 @@ function plugin_init_formcreator() {
             $PLUGIN_HOOKS['add_javascript']['formcreator'][] = 'js/scripts.js.php';
          }
 
+         if (strpos($_SERVER['REQUEST_URI'], "plugins/formcreator/front/targetticket.form.php") !== false) {
```
trasher: I guess it would be better to use `$CFG_GLPI['javascript'] `

PR #758, opened 2017-10-01
fix(ui): count of form answers in tab name

btry:
The count did not took into account the entity restriction


inc/form_answer.class.php
```diff
@@ -254,7 +254,11 @@ public function defineTabs($options = []) {
     */
    public function getTabNameForItem(CommonGLPI $item, $withtemplate=0) {
       if ($item instanceof PluginFormcreatorForm) {
-         $number  = count($this->find("`".self::$items_id."` = ".$item->getID()));
+         $dbUtils = new DbUtils();
+         $number = $dbUtils->countElementsInTableForMyEntities(
+            static::getTable(),
+            "`".self::$items_id."` = ".$item->getID()
```
orthagh: ``countElementsInTableForMyEntities`` calls ``countElementsInTable`` which calls ``$DB->request``, 
so use an array for criterion and get rid of ` char

PR #768, opened 2017-10-03
fix(ui): pqselect enabled not loaded every time it is needed

btry:
simplify code

ensure pqselect is loaded when a HTML page is loaded as well as when a tab is loaded.

Repairs the UI issue on validator user / group components

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -451,7 +451,6 @@ class PluginFormcreatorForm extends CommonDBTM
                      document.getElementById("validators_users").style.display  = "none";
                      document.getElementById("validators_groups").style.display = "none";
                   }
-                  fcInitMultiSelect();
                }
                $(document).ready(function() {changeValidators(' . $this->fields["validation_required"] . ');});';
       echo Html::scriptBlock($script);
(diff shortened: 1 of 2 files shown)


inc/form.class.php
```diff
@@ -451,7 +451,6 @@ public function showForm($ID, $options=[]) {
                      document.getElementById("validators_users").style.display  = "none";
                      document.getElementById("validators_groups").style.display = "none";
                   }
-                  fcInitMultiSelect();
```
btry: I think this is useless to init pqselect here

js/scripts.js.php
```diff
@@ -135,21 +135,24 @@ function (response) {
...
 
-   fcInitMultiSelect();
-});
-
-function fcInitMultiSelect() {
    // === Add better multi-select on form configuration validators ===
    // initialize the pqSelect widget.
+      fcInitMultiSelect();
+
    $('#tabspanel + div.ui-tabs').on("tabsload", function( event, ui ) {
```
btry: moving the event handler out of fcInitMultiSelect() because the implemetation does not fully match the name of the function.

PR #787, opened 2017-10-13
Feature/fix duplication issues

btry:


inc/form.class.php
```diff
@@ -1217,7 +1217,7 @@ public function duplicate() {
       $form_datas['name']     .= ' [' . __('Duplicate', 'formcreator') . ']';
       $form_datas['is_active'] = 0;
 
-      unset($form_datas['id']);
+      unset($form_datas['id'], $form_datas['uuid']);
```
btry: don't duplicate uuid

inc/form.class.php
```diff
@@ -1296,15 +1296,18 @@ public function duplicate() {
             return false;
          }
 
+         $new_target_item_id = $target->getField('items_id');
```
btry: get the ID of the duplicated target change or target ticket

inc/form.class.php
```diff
@@ -1296,15 +1296,18 @@ public function duplicate() {
...
             case PluginFormcreatorTargetTicket::class:
+               // Get the original target ticket
                if (!$target_ticket->getFromDB($target_values['items_id'])) {
                   return false;
                }
 
                // Update the target ticket created while cloning the target
                $update_target_ticket = $target_ticket->fields;
-               unset($update_target_ticket['id'], $update_target_ticket['uuid']);
+               $update_target_ticket['id'] = $new_target_item_id;
```
btry: this is the source target ticket, we update the ID with the ID of the duplicated one, available in the duplicated isntance of Target.

inc/form.class.php
```diff
@@ -1313,21 +1316,22 @@ public function duplicate() {
                }
 
                $new_target_ticket = new PluginFormcreatorTargetTicket();
-               $new_target_ticket->add($update_target_ticket);
-               $new_target_item_id = $new_target_ticket->getID();
-               if (!$new_target_item_id) {
+               $update_target_ticket['title'] = $update_target_ticket['name'];
+               if (!$new_target_ticket->update($update_target_ticket)) {
```
btry: this is not an add of the object, but an update, we change the method.  The prepareInputForUpdate() method of the object requires we set the title key

inc/form.class.php
```diff
@@ -1353,21 +1357,20 @@ public function duplicate() {
...
                   return false;
                }
                break;
          }
 
-         $target->update([
-            'id'        => $target->getID(),
-            'items_id'  => $new_target_item_id,
-         ]);
-
```
btry: this update does nothing becasue $target is already the duplicated instance. No need to update the items_id column

inc/form.class.php
```diff
@@ -1353,21 +1357,20 @@ public function duplicate() {
...
          switch ($target_values['itemtype']) {
             case PluginFormcreatorTargetTicket::class:
+               // Drop default actors
+               $target_ticket_actor->deleteByCriteria([
+                  'plugin_formcreator_targettickets_id' => $new_target_item_id
+               ]);
+
```
btry: whe nthe target ticket is created there are default actors added automatically. We need to drop them and create those from the source target ticket

inc/form.class.php
```diff
@@ -1381,6 +1384,11 @@ public function duplicate() {
                break;
 
             case PluginFormcreatorTargetChange::class:
+               // Drop default actors
+               $target_ticket_actor->deleteByCriteria([
+                  'plugin_formcreator_targetchanges_id' => $new_target_item_id
+               ]);
+
```
btry: the same for target changes

PR #802, opened 2017-10-18
fix(rule): location affectation on ticket via business rule

### LINKED ISSUE #795: Bug Location error on form opening
Every time the user opens a call through the form, it ignores the existing location reprieve in glpi and does not inform the called which location of the user, this test was performed in version 9.1.6 of glpi and in the latest version of plugin formcreator.

btry:
if a user has a location and submits a form the business rules to propagate location of the user
on the  generated ticket does not works

fix #795

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -1106,6 +1106,12 @@ EOS;
          } else {
             $requesters_id = $this->requesters['_users_id_requester'][0];
          }
+
+         // If only one requester, revert array of requesters into a scalar
+         // This is needed to process business rule affecting location of a ticket with the location of the user
+         if (count($this->requesters['_users_id_requester']) == 1) {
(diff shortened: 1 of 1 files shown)


btry:
validated by reporter of #795

inc/targetticket.class.php
```diff
@@ -1106,6 +1106,12 @@ public function save(PluginFormcreatorForm_Answer $formanswer) {
          } else {
             $requesters_id = $this->requesters['_users_id_requester'][0];
          }
+
+         // If only one requester, revert array of requesters into a scalar
+         // This is needed to process business rule affecting location of a ticket with the lcoation of the user
```
orthagh: lcoation -> location

btry:
Hi @mrcssilvasouza 

Please let's continue to discuss about your issue in the right place. Confirl if your isue is a duplicate if #795, and if not, show me an exact step by step way to reproduce (businiess rules included if needed). This will help me to be more efficient.

PR #803, opened 2017-10-23
dont force layout for service catalog

btry:
some tests shows that there is no harmful impact with alternate layouts.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -114,6 +114,10 @@ input.submit_button:focus {
    box-shadow: 0 0 10px rgba(233,175,102,1);
 }
 
+.layout_vsplit.form #tabspanel + div.ui-tabs:not(.horizontal) {
+   width: calc(100% - 30px) !important
+}
+
 .line0 {
    background-color: #F0F0F0;
 }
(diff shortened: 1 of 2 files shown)


btry:
see also https://github.com/glpi-project/glpi/pull/3032

css/styles.css
```diff
@@ -114,6 +114,10 @@ input.submit_button:focus {
    box-shadow: 0 0 10px rgba(233,175,102,1);
 }
 
+.layout_vsplit.form #tabspanel + div.ui-tabs:not(.horizontal) {
+   width: calc(100% - 30px) !important
```
orthagh: if you could, don't use `!important`.
A specialization selector do the same thing and can be overwritten by a more specialized selector (and not the `important` one)

btry: this will  be hard to remove !important as GLPI already uses it, and I need to override it:
https://github.com/glpi-project/glpi/blob/9.2/bugfixes/css/styles.css#L876

```css
   .layout_vsplit.form #tabspanel + div.ui-tabs:not(.horizontal) {
      width: 50% !important;
      margin: 0 15px;
      float: right;
      border: none;
      box-shadow: 0px 1px 2px 1px #D2D2D2;
   }
```

Actually, after seeing it, (presence of !important with a so complex selector) I preferred to not try to remove it :)

orthagh: ok as glpi do it, no other possibility so

PR #806, opened 2017-10-23
fix(targetticket): fix entity of generated ticket

btry:
when a form is validated the generated ticket was in the entity of the validator
when the target ticket sets the entity the current entity of the requester

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -1032,10 +1032,8 @@ EOS;
 
       $data   = [];
       $ticket  = new Ticket();
-      $form    = new PluginFormcreatorForm();
+      $form    = $formanswer->getForm();
       $answer  = new PluginFormcreatorAnswer();
-
-      $form->getFromDB($formanswer->fields['plugin_formcreator_forms_id']);
 
       // Get default request type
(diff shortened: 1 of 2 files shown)


inc/targetticket.class.php
```diff
@@ -1112,8 +1110,9 @@ public function save(PluginFormcreatorForm_Answer $formanswer) {
       switch ($this->fields['destination_entity']) {
          // Requester's entity
          case 'current' :
-            $data['entities_id'] = $_SESSION['glpiactive_entity'];
+            $data['entities_id'] = $formanswer->getField('entities_id');
```
orthagh: I 'm not ok with this modification.
'current' case, if for current active entity
form entity is 'form' case

btry: After checking in the UI : 
- the comment **Requester's entity** actually matches **case 'current'** (I  checked, just in case).
- in the UI, this settings is displayed as **active entity**. This is not clear if this is the active entity ofthe requester or the validator. In my opinion, this is  the requester's active entity. Maybe we should update  the label matching this case.
- when a requester saves a form, his current entity is saved in  the form answers. This change retrieves this entity as it reflects the active entity of the requester

PR #824, opened 2017-11-13
fix rn in generated ticket or change

btry:


btry:
The real fix goes in GLPI where a removal of \r\n is done client side with a JS callback. 

The callback runs only when the field is mandatory (CSS class required).

PR #826, opened 2017-11-13
avoid html entitization of accented chars

btry:


btry:
@orthagh I added some unit tests as suggested and handled be tter the default values, which needed some more love too.

Also I added a migration process to remove HTML entities in previously created questions.

PR #840, opened 2017-12-11
fix html tags

btry:
in rich text mode and GLPI 9.2.1, HTML tags may appear in generated tickets (followup tab)

Needs validation from reporter


btry:
Note: #840 may affect this issue.

btry:
validated by requester

PR #843, opened 2017-12-12
fix(targetticket): follow change in GLPI for due date

btry:
renamed due_date to time_to_resolve. Change in GLPI was not followed in the plugin.


btry:
validated by requester

PR #857, opened 2017-12-18
Fix inconsistency when checking question conditions

### LINKED ISSUE #829: Required condition ignored
Hi.
GLPI = 9.1.2
Forms = 2.5.2

When creating a question of:
Type: Select
Required: Yes
Show empty: Yes
Default values:
Values: *List of my values*
Description: *My description*
Hidden Unless: *Some condition*

User of the form can avoid to select some value of *List of my values*. He just left the Select field as "-----" and the required condition gets violated.

btry:
fix #829


btry:
validated by the requester

PR #858, opened 2017-12-18
WIP issues 848 854: too much encoding

btry:
Remove abusive html entities in DB

Easy job done, remains to cleanup questions table


btry:
#868 also does a part of the job

PR #862, opened 2017-12-20
Wrong  ticket display

### LINKED ISSUE #859: links of answers don't point to the expectd ticket
image

image

## steps to reproduce
1- ouvrir GLPI fresh install, créer un ticket de façon normale (sans passer par FC)
2- créer un formulaire avec une destination de type ticket et donner les droits au post-only de l'utiliser
3- activer l'interface simplifié FC dans l'entité racine
4- se connecter en post-only dans cette entité racine
5- remplir le formulaire une première fois
6- constater que le bug n'est pas reproduit (depuis la liste "My request for assistance" => url formcreator/front/issue.form.php?id=2&sub_itemtype=Ticket)
7- remplir le formulaire une deuxième fois
8- constater que le bug est reproduit (affichage du contenu du ticket id=3 pour l'items_id = 2) en consultant de nouveau l'ancien ticket (même url que le point 6)

image

vu que dans le code que j'ai pointé plus haut on considère "items_id" comme étant l'id du ticket généré.... et bien on affiche en fait le ticket id=3 :)
et évidemment, ainsi de suite dès que tu rajoutes des réponses

image

le 3 affiche le 4, le 4 affiche le 4 (vu que items_id=4 ne répond rien)...

inspect this code
```php
   public function getTicketsForDisplay($options) {
      $item = $options['_item'];
      $formanswerId = $options['id'];
      $item_ticket = new Item_Ticket();
      $rows = $item_ticket->find("`itemtype` = 'PluginFormcreatorForm_Answer'
                                  AND `items_id` = $formanswerId", "`tickets_id` ASC");

      if (count($rows) == 1) {
         // one ticket, replace item
         $ticket = array_shift($rows);
         $item = new Ticket;
         $item->getFromDB($ticket['tickets_id']);
      } else if (count($rows) > 1) {
         // multiple tickets, force ticket tab in form anser
         Session::setActiveTab('PluginFormcreatorForm_Answer', 'Ticket$1');
      }

      return $item;
   }
```

btry:
fix #859

### DIFF
--- a/inc/issue.class.php
+++ b/inc/issue.class.php
@@ -143,7 +143,9 @@ class PluginFormcreatorIssue extends CommonDBTM {
 
       // if ticket(s) exist(s), show it/them
       $options['_item'] = $item;
-      $item = $this->getTicketsForDisplay($options);
+      if ($item Instanceof PluginFormcreatorForm_Answer) {
+         $item = $this->getTicketsForDisplay($options);
+      }
 
       $item->showTabsContent();
 
@@ -191,7 +193,9 @@ class PluginFormcreatorIssue extends CommonDBTM {
 
       // retrieve associated tickets
       $options['_item'] = $item;
-      $item = $this->getTicketsForDisplay($options);
+      if ($item Instanceof PluginFormcreatorForm_Answer) {
+         $item = $this->getTicketsForDisplay($options);
+      }
 
       // force recall of ticket in layout
       $old_layout = $_SESSION['glpilayout'];


btry:
@flegastelois 

look at this screenshot
image

I apended to your scenario the use of a 2nd form which does NOT generate any  ticket. I wanted to have in my DB a answer in the list. (the one prefixes with **f_**).

If I click on f_3 the method you pointed has $item an isntance of Form_Answer. Then I deduce the code which replaces $item by a ticket instance should run only when $item is a Form_Answer only. This makes sense with the comment in the method. Also, if $item is a Ticket, then its ID is already a  tickets ID.

Can you check this patch and feedback ? I need to check the change when a form generates several tickets. Not done yet.

@orthagh I added you as rewiewer as this part was your code.

btry:
PR tested with a multi ticket form. It works (for me).

PR #905, opened 2018-02-06
fields formating in targets

btry:


inc/field.class.php
```diff
@@ -25,9 +29,22 @@ public function __construct($fields, $data = []) {
...
+      return $input;
+   }
+
+   /**
+    * Prepares a answer value for output in a target object
+    * @param string|array $input the answer to format for a target (ticket or change)
+    * @return string
+    */
+   public function prepareQuestionInputForTarget($input) {
+      return addslashes($input);
```
orthagh: Toolbox::addslashes_deep maybe ?

btry: You're right. This is safer

orthagh: apart this, it should be ok

btry: thanks !

PR #927, opened 2018-03-05
WIP use default values to set actors of tickets

### LINKED ISSUE #629: Version 2.5.0: group not added to requesters
Hello

Thx for this great new version.

Is it possible to do the same thing for ticket title as category
image
replace title by content of title of incident fields of my form

My group is not created as well 

- It would be useful to have the possibility to sort form views for users, with glpi statistics we are able to know which request are the most used and put these forms firstly for users.

btry:
This fix solves issues with Behaiours plugin < 2.0.2 and creates tickets like GLPI does internally

fix #629

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -1065,7 +1065,7 @@ EOS;
          '_groups_id_assign'           => [],
       ];
 
-      $data   = [];
+      $data = Ticket::getDefaultValues();
       $ticket  = new Ticket();
       $form    = $formanswer->getForm();
       $answer  = new PluginFormcreatorAnswer();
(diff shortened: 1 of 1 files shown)


btry:
This PR reveals an issue in GLPI : inconsistency between default values of a ticket and showForm()

inc/targetticket.class.php
```diff
@@ -1094,7 +1094,7 @@ public function save(PluginFormcreatorForm_Answer $formanswer) {
          unset($predefined_fields['_groups_id_requester']);
       }
       if (isset($predefined_fields['_groups_id_observer'])) {
-         $this->addGroupActor('assigned', $predefined_fields['_groups_id_observer']);
+         $this->addGroupActor('observer', $predefined_fields['_groups_id_observer']);
```
btry: copy paste problem => put the observer groups in assigned groups

PR #935, opened 2018-03-19
fix(condition): fix show hide issues

btry:


inc/field.class.php
```diff
@@ -81,7 +81,7 @@ public function prepareQuestionInputForTarget($input) {
    public function show($canEdit = true) {
       $required = ($canEdit && $this->fields['required']) ? ' required' : '';
 
-      echo '<div class="form-group ' . $required . '" id="form-group-field' . $this->fields['id'] . '">';
+      echo '<div class="form-group ' . $required . '" id="form-group-formcreator_field_' . $this->fields['id'] . '">';
```
btry: there is a mismatch between the ID of the tag and the key of the values returned by ajax/showdields.php

inc/fields.class.php
```diff
@@ -291,7 +291,7 @@ public static function isVisible($id, $values) {
 
       // If the field is hidden by default, show it if condition is true
       if ($question->fields['show_rule'] == 'hidden') {
-         return $return;
+         return ($return == true);
```
btry: not related: ensure isVisible returns a boolean, no 0 or 1

inc/form.class.php
```diff
@@ -991,7 +991,7 @@ class='formcreator_form form_horizontal'>";
       echo '</div>';
 
       echo '<input type="hidden" name="formcreator_form" value="' . $item->getID() . '">';
-      echo '<input type="hidden" name="_glpi_csrf_token" value="' . Session::getNewCSRFToken() . '">';
+      echo Html::hidden('_glpi_csrf_token', ['value' => Session::getNewCSRFToken()]);
```
btry: not related: ise GLPI framework when possible

js/scripts.js.php
```diff
@@ -535,7 +535,7 @@ function formcreatorChangeValueOf(field_id, value) {
    formcreatorShowFields();
 }
 function formcreatorAddValueOf(field_id, value) {
-   formcreatorQuestions[field_id] = value;
+   formcreatorQuestions['formcreator_field_' + field_id] = value;
```
btry: use **formcreator_field_{number}** everywhere (and simplify later)

btry:
validated by  the reporter

PR #938, opened 2018-03-20
Compatibility issue with internet explorer 11

### LINKED ISSUE #936: Left menu link not work with IE11
When we use IE 11, all links on left menu not works.
There is only the link on the icon that works, but not the label

No error on IE console log.

Work with other browser (Firefox, Chrome)

btry:
fix #936


stonebuzz:
Hi Btry, 

you forgot the 'saved search' label line 130

then if we reduce the menu, the label does not disappear anymore

image

Issue #662, opened 2017-07-07
multiple file uploads with a single file field type


btry:
We need to allow users to upload several files without creating several file questions. This way, we ensure the user does not reach a limit, for example he wants to upload 4 files but there are only 3 file questions.

kedson42:
Hello!
Is there any update on this?

We need to use one question to upload more than one file, the same way we can do using the File field that comes with native on GLPI.

btry:
Hi 

No update on this issue yet. This is a community request and community issues are processed at best effort.

If you want to speedup development of this feature, contact us at contact@teclib.com

kedson42:
Ah okay.

We'll try to ajust here and update the code for approval. If you're willing to give any basic instruction on what needs to be adjusted, we'll be very grateful.

Thanks

btry:
You're welcome to contribute. I don't have in mind the requirements to complete this feature, but I think reading the code in GLPI where you observe the wanted behaviour will be useful.

This issue is a feature. This means you must target the next version of Formcreator. Use the current develop branch as a base for you contribution. 

Feel free to ask if you're stuck somewhere.

kedson42:
Hi @btry , we've managed to develop this feature here.

Now the file field from formscreator gets multiple files and attach it to a ticket, also adding the files as a follow-up, the same way GLPI does natively.

We'll push the code in order to get your approval.
Thanks!

btry:
Hi

Great. I'll plan to merge it in the 2.7.0 version.

Thank you for your contribution.

lbteles:
Finally :)
https://github.com/pluginsGLPI/formcreator/pull/941

kedson42:
Awesome! 
Thanks @btry for always helpnig the community.

@lbteles , é nóis.

alyssontkd:
Great job, @lbteles !

I was in need of this functionality. Implemented here and it worked perfectly.

Thank you so much!

btry:
@alyssontkd If you open a form answer with multiple files, can you show how  this is rendered ?

By reading the code, I think we may see only the last upload. I expect the PR still needs an improvement.

kedson42:
Here how it works in our environment:

![image](https://user-images.githubusercontent.com/34751754/37767863-586ee456-2daa-11e8-922a-1d5b5d71b025.png)

![image](https://user-images.githubusercontent.com/34751754/37767886-687a00a6-2daa-11e8-9d12-2555b88ed938.png)

[...]

PR #941, opened 2018-03-21
Issue #662 - Multiple files upload

lbteles:


inc/form_answer.class.php
```diff
@@ -643,19 +643,37 @@ public function saveAnswers($data) {
...
+                       'AND' => [
+                         'plugin_formcreator_forms_answers_id' => $formanswers_id,
+                         'plugin_formcreator_questions_id'     => $questionId
+                       ]
+                     ]);
+                     $answer->update([
+                        'id'     => $answer->getID(),
+                        'answer' => $answer_value,
+                     ], 0);
+                  }
```
btry: If $answer_value is not an array, I think we can convert it onto an array then handle it with the foreach.

Check your PR is writazble by the maitnainers of the plugin, and I'll be able to push this refactor myself.

lbteles: Yeah, is writazble.

btry:
hummm.... I forgot about that : changing a commit closes a PR...

I changed the commit mesages of your 2 commits and added a new one with code refactor.. This is on your develop branch.

Open again a PR from your develop branch to run unit tests.

I also noticed a possible malfunction in the code. I'll test that next week. When you upload several files, only the last one will appear in the answer for the field, because in the foreach loop, we update the answer on each iteration. I think this update should be done after the foreach loop, and in a way which shows all uploads (maybe an implode() call).

btry:
FYI I pushed these changes in my fork : https://github.com/pluginsGLPI/formcreator/pull/943

alyssontkd:
Great job @lbteles !

TY!

lbteles:
@alyssontkd, tamo junto mano velho!

PR #943, opened 2018-03-22
Multiple file upload lbteles

btry:


btry:
Some more work:

- fixed display of form answer when multiple files were uploaded
- added migration scalar to json for id of documents in the table answers

this contr'ibutes for further code structuration on all field types.

PR #957, opened 2018-04-02
fix(question): filtering out tag question type broken

btry:
validated by the reporter

### DIFF
--- a/inc/fields.class.php
+++ b/inc/fields.class.php
@@ -76,7 +76,7 @@ class PluginFormcreatorFields
       foreach (array_keys($tab_field_types) as $field_type) {
          $classname = 'PluginFormcreator' . ucfirst($field_type) . 'Field';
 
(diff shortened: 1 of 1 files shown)


phamorais:
Hi @btry the Type of question "TAGS" was removed from FORMS 2.6.3?

btry:
Hi

No it is still there. But this type of question depends on the plugin Tags. If you have both enabled but this type of question is broken, please open an issue.

btry:
I remember I fixed a bug long ago related to this type of question when the plugin Tags is not enabled. Formcreator must check that Tags is enabled before showing such questions. Install and enable Tags and your problem shall vanish.

phamorais:
Thank you, I'll test and position you.

phamorais:
@btry after install the plugin TAG, its works! Thaaankou!

PR #969, opened 2018-04-17
Fix creation of section with abusive escape

### LINKED ISSUE #940: Problème apostrophe sur les noms de sections
Si un nom de section contient une apostrophe la génération du ticket n'est pas possible

btry:
the issue #940 was not completely solved

fix #940


btry:
validated by a bug reporter

PR #989, opened 2018-05-07
Fixed hyphen bugs and non reactive fields

Synairgis:
This request hopefully fixes all the problems linked with importing and duplicating forms with hyphens in their names. It happened because they weren't escaped properly and it messed up the underlying sql.

Also, there was a mistake in the descriptionfield.class.php. The id attribute was misconstructed which in turn made it impossible to hide or show said field.

### DIFF
--- a/inc/fields/descriptionfield.class.php
+++ b/inc/fields/descriptionfield.class.php
@@ -35,7 +35,7 @@
 class PluginFormcreatorDescriptionField extends PluginFormcreatorField
 {
    public function show($canEdit = true) {
-      echo '<div class="description_field form-group" id="form-group-field' . $this->fields['id'] . '">';
+      echo '<div class="description_field form-group" id="form-group-formcreator_field_' . $this->fields['id'] . '">';
       echo nl2br(html_entity_decode($this->fields['description']));
       echo '</div>' . PHP_EOL;
       echo Html::scriptBlock('$(function() {
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -1199,7 +1199,7 @@ class PluginFormcreatorForm extends CommonDBTM
                $data[$key] = plugin_formcreator_encode($value);
             }
          }
-
+         
          $_SESSION['formcreator']['data'] = $data;
          // Save form
       } else {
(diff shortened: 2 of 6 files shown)


btry:
Instead of using HTML entities, I would prefer single quote escaping with a method found in the class Toolbox. I'm removing HTML entities release after release.

btry:
About the Id of description field, I merged it in a dedicated commit and kept you as author. See the 2 last commits in develop branch.

Synairgis:
I didn't use html_entity_decode just for the fun of it. I used it because when I receive something like $input['name'], the apostrophes are already html_entities. This, in turn, prevents them from being properly escaped because they only revert back to their proper character when merged into an SQL query. This is why I have to decode the string with this function.

btry:
Then I must prevent HTML entities earlier in your use case. My goal is to completely drop this type of encoding, which makes the database hard to read, especially for some foreign languages like Greek or Russian.

btry:
Then, your issue partially comes from this export. In the database there were some fields which may contain HTML entities. Between 2.6.1 and 2.6.3 some of these fields are fixed, but not all. I planned to completely remove these bad encodings for 2.7.0.

btry:
FYI the export / import should be used with the same version of the plugin. I considered to make the feature work across versions, but this will make the plugin very hard to maintain. Therefore the import shall be used with an export done with the same version.

Synairgis:
Yes, I agree, but at the same time, we had to work with those exports because they're rather long (300K) and the client didn't want to completely redo the forms he created.



-----Message initial-----
De: btry <notifications@github.com>
Envoyé: mercredi 9 mai 2018 15:27
À: pluginsGLPI/formcreator <formcreator@noreply.github.com>
Cc: Tobbi Filteau <t.filteau@synairgis.com>; Author <author@noreply.github.com>
Sujet: Re: [pluginsGLPI/formcreator] Fixed hyphen bugs and non reactive fields (#989)

Then, your issue partially comes from this export. In the database there were some fields which may contain HTML entities. Between 2.6.1 and 2.6.3 some of these fields are fixed, but not all. I planned to completely remove these bad encodings for 2.7.0.

btry:
In such case you may upgrade the plugin, run the upgrade process and finally generate an export again.

btry:
This is this export which was created with the version 2.6.1, am I right ?

btry:
By reading the file, I guess they're right. 

I'll import it in a 2.6.1 version, then upgrade to 2.6.3, then export again. Next I'll test to import in a fresh installation of  the plugin and see which bugs remains. Only those remaining bus needs to be fixed.

Issue #1006, opened 2018-05-23
Problem with radio button value upon validation failure


Synairgis:
**Describe the bug**

When you create radio buttons with apostrophes in the label and then hit a validation error, upon returning to the form, your radio button selection is gone.

**To Reproduce**

1. Create a form with two fields, one radio with apostrophes in the label (ex. it's good, it's awesome) and a second text field where you'll put a validation rule (ex. max length = 5).
2. Click on one of the two radios and then enter a text too long so that the validation fails
3. Click submit. When you get taken back to the page, the radio button is no longer checked

**Expected behavior**

When the validation fails, I would expect the radio button to keep the value I previously selected. Instead, either I get the default value or nothing at all.

**Additional context**

I've located the problem inside the radiosfield.class.php. Inside the displayField() method, there's this line:

`$checked = ($this->getValue() == $value) ? ' checked' : '';`

The problem here is twofolds. You see, `$this->getValue()` returns apostrophes that are both escaped with a backslash AND in html entity form while `$value` returns the normal string with every character plainly as it is. Therefore, the "checked" condition is never true.

btry:
Hi

With the development version if the plugin I don't reproduce.

Can you tell your current version of Formcreator ?

![image](https://user-images.githubusercontent.com/14139801/40485403-b38399c8-5f5e-11e8-835a-94388beffbc7.png)

Synairgis:
As mentioned over the phone, the bug actually happens when you correct the question, because apostrophes are initially encoded as html entities... but when you correct the question, they're reverted back to normal apostrophes. For the moment, here's the patch that I applied:

`$val = stripslashes(html_entity_decode($this->getValue(), ENT_QUOTES | ENT_XML1, 'UTF-8'));`
`$checked = ($val == html_entity_decode($value, ENT_QUOTES | ENT_XML1, 'UTF-8')) ? ' checked' : '';`

btry:
Hi 

I think this issue is solved with versions 2.7 and later

btry:
Hi

I close this issue, it should  be solved now.

Issue #851, opened 2017-12-15
Problem with Greek characters glpi_plugin_formcreator_questions table


grego1981:
Values saved in glpi_plugin_formcreator_questions in columns 'name' and 'description' appear in html entities in the database.
![1](https://user-images.githubusercontent.com/9021998/34037625-35b43c68-e192-11e7-90f3-7190da613589.PNG)

btry:
Note: fix for the column name already done in #856

btry:
Hi

This issue should be solved now, like all others you reported. If such problem still occurs with latest versions, feel free to open a new ticket.

Issue #848, opened 2017-12-15
Problem with Greek characters glpi_plugin_formcreator_forms table


grego1981:
Values saved in glpi_plugin_formcreator_forms in columns 'name' and 'description' appear in html entities in the database.
![1](https://user-images.githubusercontent.com/9021998/34036612-7325889e-e18e-11e7-9fc3-b003bc3890fa.PNG)

btry:
Hi

This issue is solved with 2.7.0-beta.1. Closing

PR #1007, opened 2018-05-25
WIP Remove abusive encoding

### LINKED ISSUE #1019: It does not hide question descrition type.
It does not hide question descrition type.
Following examples screens.

screenshot-2018-6-19 chamado nti - forms - 5

screenshot-2018-6-19 chamado nti - forms - 5 1

btry:
remove all HTML entities in question definitions

see #1006 #851 #848 

closes #1019


Synairgis:
1) Create a selectbox with values "L'école" and "L'Hòpital".
2) Create a second question which should be visible when "L'école" is selected
3) Create a third question which is standard text with a max length of 5
4) Enter a text too long in the third question and submit the form

When you return to the form, the second question is hidden even though it shouldn't be.

Synairgis:
With the latest patch, I've identified two more problems with the exact same conditions as before. 

1) Just load the form and immediately, the third control gets hidden even though the value in the current list is "L'école", which means it should be visible

2) Choose "L'hòpital" as a value in the list and submit the form with a text too long in the second question. When you return to the form, the selectbox's value is back to "L'école".

btry:
The changes of the last commits introduces some regression. I'll notify you when the branch will  be stable enough for tests.

btry:
I worked hard to finish my plans to migrate unit tests from phpunit to atoum. This PR requires tests, and I want to add tests in Atoum instead of phpunit, to avoid rewrrite for Atoum (double work). 

Migration to Atoum will be merged in 2.7.0-dev, then this PR too.

btry:
Hi @Synairgis 

I added a commit which fixes the 1st scenario you described here. It seems the 2nd is solved also.

Do you want to test it and feedback ? With your double check, I may consider to merge this PR for 2.6.4.

This PR must be tested without your custom patches to avoid conflicts.

btry:
PR osbolete

PR #1047, opened 2018-07-18
Access to form not properly checked

### LINKED ISSUE #1030: Forms not restricted to an entity
Hello,

I wanted to report than when we have a form restricted in an entity, we can load it and validate it in another entity. We have simply change entity with the selector.

"Destination entity" set up to "current active entity" make than tickets are not really create where we wanted.

I'm on glpi 9.3 and formcreator 2.6.3

Best regard

btry:
Restriction by entity not done when accessing a form

### Changes description

Add check for read access to form when viewing it.

Closes #1030


btry:
Validated by the reporter; merging...

PR #1075, opened 2018-08-23
l10n – added Czech translation into xml data

p-bo:
Added short and long description as well as language code and tags in Czech language. Sorted alphabetically to correct displaying order on plugins.glpi-project.org


btry:
Hi

Thank you for your contribution.

PR #1125, opened 2018-09-25
preserve order horizontally (update masonry to 4.2.2)

orthagh:
cf: https://masonry.desandro.com/options.html#horizontalorder
This masonry option is here since 4.2.0


orthagh:
For information, tested and validated with my customer with a lot of form (50+)

PR #1129, opened 2018-09-28
Allow to use external links on the helpdesk home page

mohierf:
### Changes description

The main idea is to be able to add extra menu items in the left pane menu of the helpdesk interface. These extra items are related to Glpi external links.

I simply developed a small POC for this feature. I think that some enhancements are still necessary to make it more configuration friendly.

-----
How to:
The simplest case is to create some external links in the  Glpi DB which name is starting with `Helpdesk`. Those links will be used to create a menu entry and the `Helpdesk` keyword will be removed in the menu entry button.

Then, you can add some lines in the link comment field: 
- a line such as `Icon: fa fa-plane` allow to declare the icon to be used for the link
- a line such as `Title: the link title

### DIFF
--- a/inc/config.class.php
+++ b/inc/config.class.php
@@ -1,0 +1,175 @@
+<?php
+/*
+ * @version $Id$
+ -------------------------------------------------------------------------
+ GLPI - Gestionnaire Libre de Parc Informatique
+ Copyright (C) 2003-2011 by the INDEPNET Development Team.
+
+ http://indepnet.net/   http://glpi-project.org
+ -------------------------------------------------------------------------
+
+ LICENSE
+
+ This file is part of GLPI.
+
+ GLPI is free software; you can redistribute it and/or modify
+ it under the terms of the GNU General Public License as published by
+ the Free Software Foundation; either version 2 of the License, or
+ (at your option) any later version.
+
+ GLPI is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ --------------------------------------------------------------------------
+ */
+
+class PluginFormcreatorConfig extends CommonDBTM {
+
+   static protected $notable = true;
+
+   /**
+    * Load the plugin configuration in a global variable $PFC_CONFIG
+    *
+    * @global array $PFC_CONFIG
+    */
+   static function loadConfiguration() {
+      global $PFC_CONFIG;
+
+      // Get global configuration parameters
+      $PFC_CONFIG = Config::getConfigurationValues('formcreator');
+      if (! isset($PFC_CONFIG['user_preferences'])) {
+         Config::setConfigurationValues('formcreator', ['user_preferences' => true]);
+         $PFC_CONFIG['user_preferences'] = true;
+      }
+      if (! isset($PFC_CONFIG['avatar'])) {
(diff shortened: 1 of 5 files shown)


inc/wizard.class.php
```diff
@@ -165,6 +166,57 @@ public static function header($title) {
...
+         $label = trim(str_replace('Helpdesk', '', $a_link['name']));
+
+         /*
+          * If the link data field contains a line starting with Icon:, this line is used as a class icon
+          * for the link. As an example, this line:
+          * Icon: fa fa-dashboard
+          * will use the font awesome fa-dashboard icon for the current link.
+          * As default, the fa-globe icon is used.
+          */
+         preg_match_all("/^Icon:(.*)$/m",$a_link['data'], $found);
```
btry: Code style checks failed here, lissing a space after a comma

inc/wizard.class.php
```diff
@@ -165,6 +166,57 @@ public static function header($title) {
...
+         }
+
+         /*
+          * If the link data field contains a line starting with Title:, this line is used to find the title used
+          * for the link. As an example, this line:
+          * Title: this is my title
+          * will define a new title for the current link.
+          * As default, the link name is used.
+          */
+         preg_match_all("/^Label:(.*)$/m",$a_link['data'], $found);
```
btry: Code style checks failed here, missing a space after a comma

btry:
Hi

Your feature looks interesting; I'm wondering if we should enhance it to avoid analysis of link objects to get metadata (icon, helpdesk prefix). 

Maybe it would be better to implement a user interface to select links to use and select an icon + other useful properties.

mohierf:
Sure we should 😉  but for a simple POC to test this feature, parsing the link name was quick and easy. We should have a plugin configuration page that allows to select which links are to be exposed ... I did not see any configuration page yet else I would have implemented in the configuration form.

btry:
Hi

I'll do a functional test within  15 days (i know this is long, but I'm over-busy for a while).

Here is my suggestion for a configuration page: in the list of plugins, clicking in Form creator does the same as clicking on "Forms" in the administration menu, the user sees the list of forms.

Then the link duplicates the administration menu. We can change this to go to a configuration page related to then plugin instead of the list of forms. you at have a look to the plugin Flyve MDM to see how we can offer a complex interface with tabs even for things not related to itemtypes. The class PluginFormcreatorConfig offers the starting point for such settings interface.

mohierf:
The last commits are my current dev status:
- a try to add some configuration parameters in the main Glpi configuration
- a try to configure the items present in the top header

It is probably not the best solution to configure in the main options but it was the easiest for me to make some tests ...

mohierf:
I simply rebased my branch onto the upstream/develop and now I have many conflicting files ... 😞

btry:
Hi

this is because I recently merged a huge PR.

run the following : 
```git rebase --abort``` to abort the rebase and restore your branch to the previous state

File: inc/config.class.php [hunk 3]
```diff
@@ -0,0 +1,175 @@
...
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with GLPI. If not, see <http://www.gnu.org/licenses/>.
+ --------------------------------------------------------------------------
+ */
+
+class PluginFormcreatorConfig extends CommonDBTM {
```
btry: I would recommend a configuration relative to the current entity, not global to the instance of GLPI.

There is a file entityConfig which can contain your new settings, with inheritance support like the service catalog setting.

mohierf:
I will close this PR and open a new one includeing:
- configuration for each entity
- clean commits based upon the most recent version

btry:
Hi

OK !

Issue #1197, opened 2018-11-13
"show unless" fails on checkboxes


btry:
**Describe the bug**

issue reported in #1192 (https://github.com/pluginsGLPI/formcreator/issues/1192#issuecomment-438180954)

 
when a checkbox is displayed by a condition "show unless", the field is not hidden when the condition is not met.

Issue #1204, opened 2018-11-13
condition fail on Hostname


tdelaux:
**Describe the bug**

field Hostname does not appear when condition is OK

Issue #1203, opened 2018-11-13
condition fail on IP adress


tdelaux:
**Describe the bug**

field IP adress does not appear when condition is OK

Issue #1202, opened 2018-11-13
"show unless" fails on multiselect


tdelaux:
Describe the bug

when a multiselect is displayed by a condition "show unless", the field is not hidden when the condition is not met.

PR #1206, opened 2018-11-13
fix(QuestionCondition): inconsistency in selectors

btry:
fix application of conditions in DOM
closes: #1197 #1198 #1199 #1200 #1201 #1202 #1203 #1204

### DIFF
--- a/inc/fields.class.php
+++ b/inc/fields.class.php
@@ -268,11 +268,11 @@ class PluginFormcreatorFields
          } else {
             switch ($currentLogic) {
                case 'AND' :
-                  $return &= $value;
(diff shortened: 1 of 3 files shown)


inc/fields.class.php
```diff
@@ -332,7 +331,7 @@ public static function updateVisibility($input) {
       }
 
       $questionToShow = [];
-      foreach ($input as $id => $value) {
+      foreach ($fields as $id => $value) {
```
btry: some field types don't set a $_POST value when they are empty (example: actors field). It is better to traverse the list of questions in the form rather than the $_POST data

PR #1249, opened 2018-12-18
add test for Form::saveForm

orthagh:
no tested locally, waiting for travis :crossed_fingers:

### DIFF
--- a/tests/src/CommonTestCase.php
+++ b/tests/src/CommonTestCase.php
@@ -158,6 +158,9 @@ abstract class CommonTestCase extends CommonDBTestCase
       if (!isset($input['name'])) {
          $input['name'] = $this->getUniqueString();
       }
+      if (!isset($input['is_active'])) {
+         $input['is_active'] = true;
+      }
       $form = new \PluginFormcreatorForm();
       $form->add($input);
       $this->boolean($form->isNewItem())->isFalse();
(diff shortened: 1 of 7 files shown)


tests/src/CommonTestCase.php
```diff
@@ -158,6 +158,9 @@ protected function getForm($input = []) {
       if (!isset($input['name'])) {
          $input['name'] = $this->getUniqueString();
       }
+      if (!isset($input['is_active'])) {
+         $input['is_active'] = true;
```
btry: it should be '1', not true IMO

tests/suite-unit/PluginFormcreatorForm.php
```diff
@@ -445,4 +445,70 @@ public function testgetFormFromSection($section, $expected) {
...
+
+      // prepare input
+      $input = [
+         'formcreator_form' => $form->getID(),
+         'formcreator_field_'.$question->getID() => $answer
+      ];
+
+      // send for answer
+      $formAnswerId = $form->saveForm($input);
+      $this->boolean((bool)$formAnswerId)->isTrue();
```
btry: if a success is expected it is probably better ti check that $formAnswerId is an integer ?

(the method returns a bool only on failure)

orthagh: the cast bool do the same thing, but i can if you prefer

orthagh: this is how we do in GLPI test, we don't care to check the integer, the goal is indeed to test if we don't have a false return

btry: The following should to the job as well

```suggestion
      $this->integer($formAnswerId);
```

PR #1257, opened 2019-01-02
WIP Allow deletion of form answers

btry:
If a form has form answers then the deletion of the form is impossible (by design).

The goal is to avoid accidental deletion of those answers. The admin must delete the answers first, then delete the form.

However it it impossible to delete a form answer.

This PR allows deletion of form answers, thus unlocking deletion of their form.

@flegastelois @orthagh @Pieurre 

Do you prefer to implement a cascaded delete when one delete a form ? Or should I keep the current design which force prior deletion of form answers before deleting a form ?

### Checklist

### References

Closes #N/A

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -407,6 +407,12 @@ function plugin_formcreator_hook_restore_ticket(CommonDBTM $item) {
    }
 }
 
+/**
+ * Delete the issue linked to a ticket
+ * Called by GLPI when a ticket is about to be purged
+ *
+ * @param CommonDBTM $item
+ */
 function plugin_formcreator_hook_purge_ticket(CommonDBTM $item) {
    if ($item instanceof Ticket) {
       $id = $item->getID();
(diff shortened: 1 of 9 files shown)


hook.php
```diff
@@ -437,3 +443,56 @@ function plugin_formcreator_dynamicReport($params) {
...
+      $success = $formAnswer->deleteByCriteria([
+         PluginFormcreatorForm::getForeignKeyField() => $this->getID(),
+      ]);
+
+      if (!$success) {
+         // Failed to delete linked items
+         $item->input = false;
+      }
+   }
+}
```
btry: thus function is currently useless because of the current design (see my comment about the PR). It will be useful if we choose to do cascaded deletion.

inc/formanswer.class.php
```diff
@@ -117,6 +117,14 @@ public function canViewItem() {
       return false;
    }
 
+   public static function canPurge() {
+      if (Session::haveRight('entity', UPDATE)) {
+         return true;
+      }
+
+      return false;
+   }
```
btry: deleting form answers is allowed if the user can edit entities

btry:
oboleted by removal of is_deleted column

PR #1267, opened 2019-01-18
qtip hover added on field TITLE with the issue content

humbertocruz:
### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/issue.class.php
+++ b/inc/issue.class.php
@@ -508,6 +508,7 @@ class PluginFormcreatorIssue extends CommonDBTM {
    }
 
    public static function giveItem($itemtype, $option_id, $data, $num) {
+      $ticket = new Ticket();
       $searchopt = &Search::getOptions($itemtype);
       $table = $searchopt[$option_id]["table"];
       $field = $searchopt[$option_id]["field"];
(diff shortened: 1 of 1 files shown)


inc/issue.class.php
```diff
@@ -521,7 +522,16 @@ public static function giveItem($itemtype, $option_id, $data, $num) {
       switch ("$table.$field") {
          case "glpi_plugin_formcreator_issues.name":
             $name = $data[$num][0]['name'];
-            return "<a href='".FORMCREATOR_ROOTDOC."/front/issue.form.php?id=".$id."&sub_itemtype=".$data['raw']['sub_itemtype']."'>$name</a>";
+            $ticket->getFromDB($data[$num][0]['id']);
+            $content = $ticket->getField('content');
+            $desc = $data[$num][0]['desc'];
+            return "<a id='Ticket".$data[$num][0]['id']."' href='".FORMCREATOR_ROOTDOC."/front/issue.form.php?id=".$id."&sub_itemtype=".$data['raw']['sub_itemtype']."'>$name</a><div style='display:none' id='TicketContent".$data[$num][0]['id']."'>".nl2br($content)."</div><script type='text/javascript'>
+            //<![CDATA[
```
btry: Have a look into Html::showToolTip() . It is better to use this method instead of duplicating its feature.

inc/issue.class.php
```diff
@@ -521,7 +522,16 @@ public static function giveItem($itemtype, $option_id, $data, $num) {
       switch ("$table.$field") {
          case "glpi_plugin_formcreator_issues.name":
             $name = $data[$num][0]['name'];
-            return "<a href='".FORMCREATOR_ROOTDOC."/front/issue.form.php?id=".$id."&sub_itemtype=".$data['raw']['sub_itemtype']."'>$name</a>";
+            $ticket->getFromDB($data[$num][0]['id']);
```
btry: if the issue is from a form without ticket or with several tickets, the ID will be a PluginFormcreatorFormAnswer ID. Then you might show a qtip for the unexpected ticket.

I don't see for now a better solution than checking the 1st character if display_id (it is 't' for a ticket, or 'f' for form answer). If the character is 't' then you can show the qtip. In all other cases you should use "N/A" for the qtip, or disable it.

inc/issue.class.php
```diff
@@ -508,6 +508,7 @@ static function getDefaultSearchRequest() {
    }
 
    public static function giveItem($itemtype, $option_id, $data, $num) {
+      $ticket = new Ticket();
```
btry: You should move this line just before L525 to instanciate a Ticket only when it will be useful.

btry:
Hi @humbertocruz 

I replace this PR with #1305 to add some improvements

PR #1275, opened 2019-01-23
More resilient export / import

btry:
If many forms are being exported, it may be possible to hit the php time limit.

fixes also warnings on import when the export of form_profile is incomplete
### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -1914,7 +1914,6 @@ class PluginFormcreatorForm extends CommonDBTM implements PluginFormcreatorExpor
 
          $importLinker = new PluginFormcreatorImportLinker();
          foreach ($forms_toimport['forms'] as $form) {
-            //self::import($form);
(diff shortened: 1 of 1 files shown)


inc/form_profile.class.php
```diff
@@ -149,6 +149,9 @@ public static function import($forms_id = 0, $form_profile = []) {
       $formFk = PluginFormcreatorForm::getForeignKeyField();
       $form_profile[$formFk] = $forms_id;
 
+      if (!isset($form_profile['_profile']) || $form_profile['_profile'] === null) {
+         return false;
```
btry: Should we fail or should we show a warning ?

@orthagh what do you think about this kind of case (will be seen also for all itemtypes in the core

btry:
I must remove the part about the profiles : a form can have no profile, then the code is faulty in those cases.

PR #1289, opened 2019-02-05
escape form import

orthagh:
Question, is an import test exists ?


btry:
Hi

No, there is no unit tests for import yet. I think that import feature needs dedicated classes to make unit tests easier to write.

PR #1294, opened 2019-02-06
Escaping for target

btry:
### Changes description

add more escaping for target ticket and target change

Also adds a code factorization, and a comment correction

### Checklist

### References

Closes #N/A


btry:
@orthagh among other fix and refactor, I found there is a need to escape one more field for target ticket, and escape the same fields for target change.

PR #1300, opened 2019-02-12
Category not assigned

btry:
also refactor entity assignation for methods consistency


inc/targetbase.class.php
```diff
@@ -297,7 +299,36 @@ protected function getTargetEntity(PluginFormcreatorFormAnswer $formanswer, $req
             break;
       }
 
-      return $entityId;
+      $data[$entityFk] = $entityId;
+      return $data;
+   }
+
+   protected function setTargetCategory($data, $formanswer) {
```
btry: method moved from Target Ticket to share with Target Change

PR #1311, opened 2019-02-21
Remove enum from DB

btry:
### Changes description
GLPI does not uses enum types, then we convert them into integers

### Checklist

### References

Closes #N/A


inc/formanswer.class.php
```diff
@@ -41,6 +41,12 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    const SOPTION_ANSWER = 900000;
 
+   // Values choosen to not conflict with status of ticket constants
+   // @see PluginFormcreatorIssue::getNewStatusArray
```
btry: Need harmonization with GLPI

ideas : 
- have similar constants in CommonITILObject
- convert formanswers into something inherited from ComonITILObject (seems complicated) ?

btry:
replaced by #1317 : cleaner commits while fixing issues

PR #1322, opened 2019-02-26
add ticket type setting to target  ticket

btry:
### Changes description

Remove the relation between a target ticket and a ticket template.

Add a ticket type for request / incident

Migrate ticket type and predefined actors to the target ticket

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -35,6 +35,13 @@ if (!defined('GLPI_ROOT')) {
 
 class PluginFormcreatorTargetTicket extends PluginFormcreatorTargetBase
 {
+   public static function getEnumTicketType() {
+      return [
+         Ticket::INCIDENT_TYPE  => __('Incident', 'formcreator'),
+         Ticket::DEMAND_TYPE   => __('Request', 'formcreator'),
+      ];
+   }
+
    public static function getTypeName($nb = 1) {
       return _n('Target ticket', 'Target tickets', $nb, 'formcreator');
    }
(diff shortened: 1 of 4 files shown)


inc/targetticket.class.php
```diff
@@ -35,6 +35,17 @@
 
 class PluginFormcreatorTargetTicket extends PluginFormcreatorTargetBase
 {
+
+   const TICKET_TYPE_INCIDENT = 1;
+   const TICKET_TYPE_REQUEST = 2;
+
+   public static function getEnumTicketType() {
+      return [
+         PluginFormcreatorTargetTicket::TICKET_TYPE_INCIDENT  => __('Incident', 'formcreator'),
```
orthagh: Why not reusing the constants of Ticket class ?

orthagh:
BTW, ticket's template also give predefined ticket's tasks so remove the relation is not a good idea

btry:
> BTW, ticket's template also give predefined ticket's tasks so remove the relation is not a good idea.
Tickets's documents also

Only the type of ticket and actors are retrieved by the plugin. Only the ticket type is generally used because the actors are handled by the plugin. Removing the support of template helps to enhance the UX by configuring all settings from the plugin. Unfortunately the ticket templates don't allow everything permitted by the plugin (example : link tickets, add several actors of the same type).

inc/targetticket.class.php
btry: changed

orthagh:
I insist, we currently have customers using the **task**'s template by a **ticket**'s template predefined in form. So no, `type` is not the only remaining field set by formcreator.

Also, linked documents and hidden fields on final ticket

btry:
OK, then I'll change the PR to provide the ticket type without need of any template, and keep the template association.

inc/targetticket.class.php
```diff
@@ -541,6 +573,38 @@ public function pre_deleteItem() {
...
+         unset($predefined_fields['_groups_id_observer']);
+      }
+      if (isset($predefined_fields['_groups_id_assign'])) {
+         $this->addGroupActor('assigned', $predefined_fields['_groups_id_assign']);
+         unset($predefined_fields['_groups_id_assign']);
+      }
+      $data = array_merge($data, $predefined_fields);
+
+   }
+
```
btry: moved from save()

PR #1328, opened 2019-02-28
fix(form): fix anonymous file upload

### LINKED ISSUE #925: Attachments. If the user is not logged in to GLPI the attachment does not go up.
We're having trouble with the form attachments. If the user is not logged in to GLPI the attachment does not go up.

Version 2.6.2

btry:
### Changes description

A fake session is ceated by  the plugin to run an anonymous form, but it is not saved server side, making the file upload fail.

### Checklist

### References

Closes #925

### DIFF
--- a/inc/fields/filefield.class.php
+++ b/inc/fields/filefield.class.php
@@ -43,15 +43,11 @@ class PluginFormcreatorFileField extends PluginFormcreatorField
       if ($canEdit) {
          $required = $this->isRequired() ? ' required' : '';
 
-         echo '<input type="hidden" class="form-control"
-                  name="formcreator_field_' . $this->fields['id'] . '" value="" />' . PHP_EOL;
-
          echo Html::file([
             'name'    => 'formcreator_field_' . $this->fields['id'],
             'display' => false,
             'multiple' => 'multiple',
          ]);
-
       } else {
          $doc = new Document();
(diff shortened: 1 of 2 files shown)


inc/fields/filefield.class.php
```diff
@@ -43,15 +43,11 @@ public function displayField($canEdit = true) {
       if ($canEdit) {
          $required = $this->isRequired() ? ' required' : '';
 
-         echo '<input type="hidden" class="form-control"
```
orthagh: Are you sure about this deletion ?

btry: Yes : I think that I forgot to remove this line when I removed the HTML file upload with the version provided by Html::file() . several things confirm that : 
- the name of this input is the same as the name of the input generated by Html::file()
- when GLPI uses Html::file() it does not creates an extra input tag outside of the method

btry: yes : this input is useless

v2.6.0 before using Html::file()
https://github.com/pluginsGLPI/formcreator/blob/2.6.0/inc/fields/filefield.class.php#L8

v2.6.1 after using Html::file()
https://github.com/pluginsGLPI/formcreator/blob/2.6.1/inc/fields/filefield.class.php#L8

the input has always an empty value.

moreover the input following it has the same name.

PR #1337, opened 2019-03-04
fix Import of entity and category

btry:
- avoid import or update of forms in an entity where the user does not has right for update
- fix import of form category

### Changes description
If the imported form exists and is in an unmodifiable entity, abort and warn the user

image

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -1734,12 +1734,12 @@ class PluginFormcreatorForm extends CommonDBTM implements PluginFormcreatorExpor
       // replace dropdown ids
       if ($form['plugin_formcreator_categories_id'] > 0) {
          $form['_plugin_formcreator_category']
-            = Dropdown::getDropdownName('glpi_plugin_formcreator_categories',
+            = Dropdown::getDropdownName(PluginFormcreatorCategory::getTable(),
                                         $form['plugin_formcreator_categories_id']);
       }
       if ($form['entities_id'] > 0) {
          $form['_entity']
-            = Dropdown::getDropdownName('glpi_entities',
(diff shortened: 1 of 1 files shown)


inc/form.class.php
```diff
@@ -2028,34 +2028,57 @@ public static function import(PluginFormcreatorImportLinker $importLinker, $form
...
 
       $form_obj = new self;
-      $entity   = new Entity;
-      $form_cat = new PluginFormcreatorCategory;
+
+      $forms_id = plugin_formcreator_getFromDBByField(
+         $form_obj,
+         'uuid',
+         $form['uuid']
+      );
```
btry: this code was later in the flow. Moved here to detect early if we will update an existing form for entity check needs.

inc/form.class.php
```diff
@@ -2028,34 +2028,57 @@ public static function import(PluginFormcreatorImportLinker $importLinker, $form
...
-                                                   $form['_plugin_formcreator_category'])) {
-         $form['plugin_formcreator_categories_id'] = 0;
+      $form[$entityFk] = $entityId;
+
+      $formCategory = new PluginFormcreatorCategory();
+      $formCategoryFk = PluginFormcreatorCategory::getForeignKeyField();
+      $formCategoryId = 0;
+      if (isset($form['_plugin_formcreator_category'])) {
+         $formCategoryId = $formCategory->import([
+            'completename' => $form['_plugin_formcreator_category'],
```
btry: plugin_formcreator_getFromDBByField will get the category if it exists. If not we need to create it,  then let's use CommonTreeDropdown::import() which handles both operations.

PR #1390, opened 2019-04-18
Uploaded files lost

### LINKED ISSUE #1367: Form does not insert attachments in ticket
I have GLPI 9.4.4.1 and formcreator 2.80 and when I try to open a ticket through the form the attachment is not included, I ran some tests and the attachment is included only when there is only the files field, if any other add the ticket is opened without the file.

example dont attach
image
image

example attach
image
image

btry:
When a requester uploads a file then trigger a field visibility update, files are processed early and moved.
When saving answers, the file is no longer available for building the answer and the generated targets

Closes #1367 #1358


btry:
Got positive feedback, merging...

PR #1428, opened 2019-06-18
Search items dropdownfield

btry:
Improve the selection of a glpi object. When it has a serial or other_serial field, then the plugin makes the dropdown searchable with those columns.


btry:
validated. To merge soon

PR #1432, opened 2019-06-19
Import of form category with single quote

btry:
Fix import of category with a single quote


btry:
validated by reporter

PR #1437, opened 2019-06-21
Restrict available LDAP servers.

cosmedd:
Hi!

At work, we found a issue with your plugin. We realized that a new client that started using our GLPI could see and use any LDAP configured through the "LDAP Select" type.

This means that any client, with enough privileges in his entity, can see/use our LDAP and any LDAP of other clients.

We quickly implement an workaround based on many assumptions. There is probably a better way to fix this problem but I provide this pull request in case it helps someone or if you want to use it as a starting point.

### Changes description

You can view any available LDAP if you have access
to the root entity.

Otherwise, only the LDAP configured as the default
for the current entity is displayed.

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -836,8 +836,15 @@ class PluginFormcreatorQuestion extends CommonDBChild implements PluginFormcreat
       if ($ldap_values === null) {
          $ldap_values = [];
       }
+      $current_entity = $_SESSION['glpiactive_entity'];
+      if ($current_entity == 0) {
+         $auth_ldap_condition = '';
+      } else {
+         $auth_ldap_condition = "glpi_authldaps.id = (select glpi_entities.authldaps_id from glpi_entities where id=${current_entity})";
+      }
       Dropdown::show('AuthLDAP', [
          'name'      => 'ldap_auth',
+         'condition' => $auth_ldap_condition,
          'rand'      => $rand,
          'value'     => (isset($ldap_values['ldap_auth'])) ? $ldap_values['ldap_auth'] : '',
          'on_change' => 'change_LDAP(this)',


inc/question.class.php
```diff
@@ -836,10 +836,17 @@ public function showForm($ID, $options = []) {
       if ($ldap_values === null) {
          $ldap_values = [];
       }
+      $current_entity = $_SESSION['glpiactive_entity'];
+      if ($current_entity == 0) {
+        $auth_ldap_condition = '';
+      } else {
+        $auth_ldap_condition = "glpi_authldaps.id in (select a.id from glpi_entities as e ,glpi_authldaps as a where a.id=e.authldaps_id and e.id=${current_entity})";
```
btry: Your condition seems over-complicated. Why not use the following instead ?

```'glpi_authldaps.id = (select glpi_entities.authldaps_id from glpi_entities where id=${current_entity})```

File: inc/question.class.php [hunk 2]
```diff
@@ -836,10 +836,17 @@ public function showForm($ID, $options = []) {
...
+      }
       Dropdown::show('AuthLDAP', [
          'name'      => 'ldap_auth',
+         'condition' => $auth_ldap_condition,
          'rand'      => $rand,
-         'value'     => (isset($ldap_values['ldap_auth'])) ? $ldap_values['ldap_auth'] : '',
+         'value'     => '',
```
btry: The field must show the value currently stored in DB. If one edit a question it will need to set again the LDAP to use. This might create headache to users.

cosmedd:
Requested changes addressed.

Thanks for the review, @btry!

cosmedd:
Well, I do not know if I should do something... 

Github keeps showing "Changes requested" and "Merging can be performed automatically once the requested changes are addressed."

btry:
Hi

This is normal that the status remains "changes requested". I need to review your changes again, then update the status. Please be patient until I review the PR again.

PR #1457, opened 2019-07-29
feat(ldapfield): comparisons support

### LINKED ISSUE #1454: hidden unless does not work if condition is other question with LDAP Select 2.8.3
hidden unless does not work if condition is other question with LDAP Select with plugin version 2.8.3 glpi version 9.4.3

**To Reproduce**
1. make question 
type = LDAP Select, 
filter = (& (samaccountname=*) (& (displayName=*) (&(objectCategory=person)(objectClass=user)(memberOf:1.2.840.113556.1.4.1941:= cn=DG MP All Staff, OU=SharedMailboxes,OU=Moscow, DC=mpmsk, DC=corp)(givenName=*)(sn=*))))
Attribute = (LDAP) City
2. make second question
show field = hidden unless
1st question = (in my case for ex) Moscow
3. try the form

expecting when in 1st question choose Moscow 2nd question should appear - but it does not.

**Screenshots**
image
image
image
image

**Desktop (please complete the following information):**
 - OS: Win10
 - Browser chrome

it workes in plugin version 2.5.2 glpi ver 9.2.2

btry:
fix #1454

### DIFF
--- a/inc/fields/ldapselectfield.class.php
+++ b/inc/fields/ldapselectfield.class.php
@@ -199,19 +199,19 @@ class PluginFormcreatorLdapselectField extends PluginFormcreatorSelectField
    }
 
    public function equals($value) {
-      throw new PluginFormcreatorComparisonException('Meaningless comparison');
+      return $this->value == $value;
    }
 
    public function notEquals($value) {
-      throw new PluginFormcreatorComparisonException('Meaningless comparison');
+      return !$this->equals($value);
    }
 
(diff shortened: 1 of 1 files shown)


achekmez:
Thank you, it helped. Strange thet noone uses it, and even you said the it was unexpected to use it. But it is very conviniant to get lists of cities, offices, departaments, positions online from AD and make condition upon it. Hope in next releases it will be working. 
Thank you once more.

PR #1463, opened 2019-08-02
Allow object properties in placeholders

AdrienClairembault:
Internal ref: 17110

Allow to access objects properties in placeholder, e.g. `##answer_2.Login##` through their searchOptions.
The properties name must match the name of one of the object searchOption with underscore instead of spaces.

For example, to access the value of the "Default entity" search option, you must use:
`##answer_2.Default_entity##`


inc/targetbase.class.php
```diff
@@ -1049,112 +1049,124 @@ protected function parseTags($content, PluginFormcreatorFormAnswer $formanswer,
...
+            );
+
+            return $content;
          }
+      } finally {
+         // Put the old locales on succes or if an expection was thrown
+         $TRANSLATE->setLocale($oldLocale);
       }
-
-      return $content;
```
btry: if we catch an exception, the method will not return anything. Either we need to throw an exception (and catch it in the caller), either we should return a value.

inc/targetbase.class.php
```diff
@@ -1048,11 +1048,123 @@ protected function parseTags($content, PluginFormcreatorFormAnswer $formanswer,
                }
             }
          }
+
+         if ($question->getField('fieldtype') == 'glpiselect') {
+            $content = $this->parseObjectProperties(
+               $question,
+               $questionId,
```
btry: This is the ID of the questoin given in 1st argument. I think this is useless to pass it here.

btry:
To port this feature in Formcreator 2.9.0 (branch support/2.9.0)

you may move all the new code in PluginFormcreatorDropdownField (ancestor of PluginFormcreatorGlpiselectField). Your new method parseObjectProperties() would be protected. In it you need some properties of the question. When a a PluginFormcreatorXXXXField is instanciated, it has a property named $fields which is populated with the properties of the matching question (https://github.com/pluginsGLPI/formcreator/blob/support/2.9.0/inc/field.class.php#L52)

This makes the arguments $question and $questionID useless as all the needed data are available in $this->fields

PR #1485, opened 2019-08-30
Allow self-service to see formcreator objects

AdrienClairembault:
This is needed to access the formcreator objects though the API as a self-service user.

### DIFF
--- a/inc/category.class.php
+++ b/inc/category.class.php
@@ -35,12 +35,12 @@ if (!defined('GLPI_ROOT')) {
 
 class PluginFormcreatorCategory extends CommonTreeDropdown
 {
-   static function canView() {
-      return Session::haveRight('ticket', CREATE);
-   }
-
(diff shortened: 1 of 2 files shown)


inc/category.class.php
```diff
@@ -37,6 +37,10 @@
 
 class PluginFormcreatorCategory extends CommonTreeDropdown
 {
+   static function canView() {
+      return Session::haveRight('ticket', CREATE);
```
btry: rightname of a dropdown is "dropdown" (defined in CommonDropdown)

If you redefined this method this way, you may break access to the list of categories for administrators.

I think he following will be better

```suggestion
      return parent::canView() || $_SESSION['glpiactiveprofile']['interface'] == 'helpdesk';
```

AdrienClairembault:
I've made the changes you requested and rebased on 2.9.

Issue #1486, opened 2019-08-30
Submit form through API


AdrienClairembault:
Unless I'm mistaken, it is not possible to submit a form through the API.

What would be the best way to allow this ?

A quick solution would be to create a new `PluginFormcreatorFormSubmit` object that would call `$form->saveForm($_POST)` inside it's `add` method but this feels a bit hacky.

Do you have any better idea ?

btry:
Hi

The plugin needs some code refactor :
when a user submits a form, we enter in PluginFormcreatorForm::saveForm() where the inputs from the user are validated before calling PluginFormcreatorFormAnswer::saveAnswers()

PluginFormcreatorForm::saveForm() should be (roughly) migrated and refactored in PluginFormcreatorFormAnswer, maybe in PrepareInputrForAdd().

AdrienClairembault:
Closing this as this is now working after your latest changes.

PR #1495, opened 2019-09-06
Allow forms to be submitted through API

AdrienClairembault:
As seen in #1486, the $form->saveForm() endpoint is not reachable from the API.

This is a simple fix to allow form creation though the API using a new "PluginFormcreatorFormAnswerAPI" endpoint that forward the request to the $form->saveForm() endpoint.  

As we discussed, the correct way to solve this would be use the $formanswer->add() entry point but it would require a lot of code re-factorization which require a good knowledge of this plugin and time that I currently lake.  

These proposed changes won't impact the plugin in any way and allow us to move forward with this feature (the "clean" way can always be implemented later on).

### DIFF
--- a/inc/formanswerapi.class.php
+++ b/inc/formanswerapi.class.php
@@ -1,0 +1,56 @@
+<?php
+/**
+ * ---------------------------------------------------------------------
+ * Formcreator is a plugin which allows creation of custom forms of
+ * easy access.
+ * ---------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


btry:
Hi 

I think this is not a good solution. Next week I'll work on a few fixes for the branch support/2.9.0. I'll search how to refactor the code the way I have in mind to avoid such hack. I'll notify you once done.

PR #1500, opened 2019-09-12
Parse properties only for PluginFormcreatorGlpiselectField

AdrienClairembault:
There was a little mistake when porting this feature to 2.9, it should only be enabled for GlpiselectField (and not all dropdown).


AdrienClairembault:
Rebase seems to have fixed the tests.

Issue #1483, opened 2019-08-28
Get QuestionRegex from API


AdrienClairembault:
**Describe the bug**
`GET /apirest.php/PluginFormcreatorQuestionRegex` fail with  `500 Internal Server Error`.

Server logs:
```
2019/08/28 15:02:43 [error] 1206#1206: *1278 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught ArgumentCountError: Too few arguments to function PluginFormcreatorQuestionParameter::__construct(), 0 passed in /home/teclib/Main/html/glpi-fusion-94/inc/api.class.php on line 1139 and exactly 2 expected in /home/teclib/Main/html/glpi-fusion-94/plugins/formcreator/inc/questionparameter.class.php:59
Stack trace:
#0 /home/teclib/Main/html/glpi-fusion-94/inc/api.class.php(1139): PluginFormcreatorQuestionParameter->__construct()
#1 /home/teclib/Main/html/glpi-fusion-94/inc/apirest.class.php(237): API->getItems('PluginFormcreat...', Array, 0)
#2 /home/teclib/Main/html/glpi-fusion-94/apirest.php(48): APIRest->call()
#3 {main}
  thrown in /home/teclib/Main/html/glpi-fusion-94/plugins/formcreator/inc/questionparameter.class.php on line 59" while reading response header from upstream, client: 127.0.0.1, server: *.glpi-fusion-94.local, request: "GET /apirest.php/PluginFormcreatorQuestion/4/PluginFormcreatorQuestionRegex HTTP/1.1", upstream: "fastcgi://unix:/run/php/php7.2-fpm.sock:", host: "72.glpi-fusion-94.local"
```

**To Reproduce**
Make a `GET` request on `/apirest.php/PluginFormcreatorQuestionRegex`

**Expected behavior**
Response: Collection of PluginFormcreatorQuestionRegex

**Additional context**
I'm trying to use formcreator through the API as a self-service user.
I've already fixed some rights issues (will do a PR later) but this problem seems a bit more complex.

btry:
Hi

This is vert likely causes by a HTTP PUT or POST with inconsistent data. I'll change this to have a more friendly HTTP error code

btry:
Hi

You may implment  PluginFormcreatorQuestion::post_getItem() to automatically add some properties when the object is retrieved from he DB

[...]

PR #1503, opened 2019-09-23
Get full form data from one API call

AdrienClairembault:
### Changes description

Allow to get full form data (sections, questions, conditions, ranges and regexes) in one API request.
Solve #1483.

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -1186,4 +1186,101 @@ PluginFormcreatorDuplicatableInterface
       $items = $question->getQuestionsFromFormBySection($formId, $crit);
       Dropdown::showFromArray($name, $items, []);
    }
+
+   /**
+    * Get linked data (conditions, regexes or ranges) for a question
+    *
+    * @param string    $table   target table containing the needed data (
+    *                           condition, range or regex)
+    * @param int|array $id      a single id or an array of ids
+    * @return array
+    */
+   public static function getQuestionDataById($table, $id) {
+      global $DB;
+
+      $validTargets = [
+         \PluginFormcreatorQuestion_Condition::getTable(),
+         \PluginFormcreatorQuestionRegex::getTable(),
+         \PluginFormcreatorQuestionRange::getTable(),
+      ];
+
+      if (array_search($table, $validTargets) === false) {
+         throw new \InvalidArgumentException("Invalid target ('$table')");
+      }
+
+      return iterator_to_array($DB->request([
+         'FROM' => $table,
+         'WHERE' => [
+            "plugin_formcreator_questions_id" => $id
+         ]
+      ]));
+   }
+
+   /**
+    * Get either:
+    *  - questions, conditions, regexes and range of target parent sections
+    *  - conditions, regexes and range of target question
+    *
+    * @param int $parents target parent sections
+    * @param int $id target question
+    * @return array
+    */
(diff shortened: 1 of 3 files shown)


inc/question.class.php
```diff
@@ -1186,4 +1186,93 @@ public static function dropdownForForm($formId, $crit, $name, $value) {
       $items = $question->getQuestionsFromFormBySection($formId, $crit);
       Dropdown::showFromArray($name, $items, []);
    }
+
+   /**
+    * Get target subdata (conditions, regexes or ranges)
+    *
+    * @param string $target target table
+    * @param id|array $condition condition
```
btry: @param integer|array ...

inc/question.class.php
```diff
@@ -1186,4 +1186,93 @@ public static function dropdownForForm($formId, $crit, $name, $value) {
...
+      global $DB;
+
+      $validTargets = [
+         \PluginFormcreatorQuestion_Condition::getTable(),
+         \PluginFormcreatorQuestionRegex::getTable(),
+         \PluginFormcreatorQuestionRange::getTable(),
+      ];
+
+      if (array_search($target, $validTargets) === false) {
+         throw new \InvalidArgumentException("Invalid target ('$target')");
```
btry: need a try / catch somewhere to avoid an Error 500 if this exception is triggered

inc/section.class.php
```diff
@@ -383,4 +383,49 @@ public function showSubForm($ID) {
...
+         $ids = [];
+         foreach ($data['_sections'] as $section) {
+            $ids[] = $section['id'];
+         }
+
+         $data = $data + \PluginFormcreatorQuestion::getFullData($ids);
+      } else {
+         if ($id == null) {
+            throw new \InvalidArgumentException(
+               "Parameter 'id' can't be null if parameter 'parent' is not specified"
```
btry: Same as above : if you throw an exception, you need to have a try / catch in one of the caller methods

inc/question.class.php
```diff
@@ -1186,4 +1186,93 @@ public static function dropdownForForm($formId, $crit, $name, $value) {
...
+      }
+
+      return iterator_to_array($DB->request([
+         'FROM' => $target,
+         'WHERE' => [
+            "plugin_formcreator_questions_id" => $condition
```
btry: A suggestion to have a more powerful code : maybe $condition should be an array of the complete WHERE statement of the query. See CommonDBTM::getFromDBByCrit().

I would write the array of the request as the following. 
```php
      return iterator_to_array($DB->request([
         'FROM' => $target,
         'WHERE' => $condition
      ]));  
```

and adapt all calls of the method. This makes the name $condition more appropriate also.

inc/question.class.php
AdrienClairembault: This kind of exceptions should never be caught (or would need to be caught in a global try catch at the top level of the framework but this doesn't exist in GLPI). This exception should never occur in a correct implementation, It therefore does not make sense to try to catch and handle it. Instead, the code that triggered the exception should be fixed.

inc/question.class.php
AdrienClairembault: I think the method was badly named, it is only used to avoid repetitive queries in getFullData (and thus I do not wish to rewrite the full condition), I've made some changes and it should be more straightforward now.

PR #1576, opened 2019-11-27
Fix form link for formcreator helpdesk

AdrienClairembault:
Internal ref: 18557.

### DIFF
--- a/js/scripts.js.php
+++ b/js/scripts.js.php
@@ -31,6 +31,8 @@
 
 include ('../../../inc/includes.php');
 header('Content-Type: text/javascript');
+
+$form_link_target = plugin_formcreator_replaceHelpdesk() ? "wizard.php" : "formlist.php";
 ?>
 
 var modalWindow;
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
ping @btry, can you validate this so I can close the internal ref ?

btry:
Hi

I would prefer an other solution with a redirection : I removed as many php code inclusion as possible. The goal is to have a plain JS file without dynamic content.

I believe you can get the same result by adding a redirection in setup.php, function plugin_formcreator_init() . There are already some redirections here to maintain the user in the service catalog.

AdrienClairembault:
In this case we do not wish to have a permanent redirect as this page should still be reachable by the user (we just don't want it to be the target of this specific link).

btry:
Yes, and  there are already redirects done when some conditins are met. Have a look here: 

https://github.com/pluginsGLPI/formcreator/blob/develop/setup.php#L142

AdrienClairembault:
This does not seem convenient to me as we only want to modify a single link (and not every request to a page). It should be possible with redirects by using http_referer but quite ugly compared to a dynamic link which is functionally what we want.

I'll resolve the internal ref and let you implement this the way you want since I am not sure how to do it this way.

PS: As a side note, while "dynamic" mixed js/php files are definitely bad, my personal opinion is that it would seem strange to delegate clear js features to the server code through "fixes" like redirection. You can get rid of most of the disadvantages of theses mixed files by dynamically setting some JS vars before including your now plain js files.

PR #1599, opened 2020-01-06
Purge message if answers exist

### LINKED ISSUE #1589: Unable to delete a form (v 2.8.6)
**Describe the bug**
I want to delete a form that is in the trash and I have a blocking error message. I cannot delete this form. I am super-user other old forms are abble to supress without problem.
Message in red at lower right :
Erreur
Opération échouée
(1 problèmes de droits, 0 échecs)
Saisie de formulaire xxxx - test (1) : Erreur d'habilitation

v2.8.6

btry:
When a form has answers, it is impossible to porge it. The admin must porge the answers first but no message tells it.

fix #1589

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -570,11 +570,21 @@ PluginFormcreatorDuplicatableInterface
       echo '</td>';
       echo '</tr>';
 
+      echo '<tr>';
       echo '<td>'.__('Default form in service catalog', 'formcreator').'</td>';
       echo '<td>';
       Dropdown::showYesNo("is_default", $this->fields["is_default"]);
       echo '</td>';
       echo '</tr>';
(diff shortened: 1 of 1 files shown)


mab18:
Hi !
Happy new year ;-)
It's a good option : we know why it's not working. I tested it ok.
What to do to translate the text ?

btry:
Happy new year

Translations are hosted on Transifex. 

I need to update the source strings, then contributors can translate the strings, and use the development tools (composer, robo.li) to download locales and build the .mo files for the plugin  . Usually this is done when I release a new version.

This string is not for end users then you should consider to wait the next version of  the plugin.

mab18:
Thanks for the answer ;-)

PR #1667, opened 2020-02-06
extra searchoptions for FormAnswers

btry:
The itemtype FormAnswers shall include extra search options when a user shows a list of forn answers for a given form instance.

The form to edit display prefs never has the required session var to add the specific search options.


inc/answer.class.php
```diff
@@ -65,4 +65,26 @@ public static function canView() {
...
+
+   /**
+    * Define how to display a specific value in search result table
+    *
+    * @param  string $field   Name of the field as define in $this->getSearchOptions()
+    * @param  mixed  $values  The value as it is stored in DB
+    * @param  array  $options Options (optional)
+    * @return mixed           Value to be displayed
+    */
+   public static function getSpecificValueToDisplay($field, $values, array $options = []) {
```
btry: @orthagh : I requested your review because of this method. I don't have a better way to properly format some field types. This implementation may have low performance on big lists, but the format depends on metadata of the field type.

How it works ? the method does not has any argument providing a question ID. Then to determine the field type I use a search option providing the ID of the answer instead of the answer itself.

The row in the DB contains the question ID and give access to the actual field type. Then I use the methods dedicated to output the answer in a readable format.

inc/fieldinterface.class.php
```diff
@@ -100,105 +110,116 @@ public function getValueForDesign();
...
     * @return boolean true on sucess, false otherwise
     */
    public function parseAnswerValues($input, $nonDestructive = false);
 
    /**
-    * Prepares an answer value for output in a target object
-    * @param  string|array $input the answer to format for a target (ticket or change)
-    * @return string
-    */
-   //public function prepareQuestionInputForTarget($input);
```
btry: previously obsoleted, cleaning up the interface

inc/formanswer.class.php
```diff
@@ -1551,4 +1560,33 @@ public static function getMyLastAnswersAsValidator($limit = 5) {
...
+      if (!isset($_SESSION['formcreator']['form_search_answers'])) {
+         return false;
+      }
+
+      if (strpos($_SERVER['REQUEST_URI'], $CFG_GLPI['root_doc'] . '/ajax/common.tabs.php') !== 0) {
+         return false;
+      }
+
+      if (isset($_GET['displaytype']) && $_GET['displaytype'] == PluginFormcreatorFormAnswer::class) {
+         // return true;
```
btry: enabling this gives a way to display or not the fields specific to the form. Not sure this is needed (may break if a form has 10 fields and display prefs are configured to show all of them, but an other has only 5)

inc/answer.class.php
orthagh: example:
https://github.com/glpi-project/glpi/blob/9.4/bugfixes/inc/commonitilobject.class.php#L2670-L2693

this method is always called for the current object (and the corresponding table), so you just need to filter on the field

btry: The method belongs to PluginFormcreatorAnswer, and the context is static then I cannot use an instance to find the PluginFormcreatorQuestion object containing the type of field. 
I have several cases 
- a string to be displayed as is
- a number which is an ID of an other itemtype (glpi object, dropdown fields)
- a number to convert into a string from an array (urgency field)
- a set of values separated by \r\n (checkboxes, multiselecti fields)

Only the 2 first can be processed without this method. The 2 last cases needs a different processing which require that I can determine the type of field stored in the associated PluginFormcreatorQuestion instance.

By the way, maybe should I protect the trnasformation by testing the column name.

btry:
obsolete

PR #1680, opened 2020-02-13
fix(dropdownfield,glpiselectfield): empty value parameter not honored

btry:
With entity itemtype, a dropdown does not show the empty value.

Also itemtypes having serial or otherserial don't respect show_empty parameter.


btry:
Validated by a user

PR #1681, opened 2020-02-13
Last valid answer

AdrienClairembault:
### Changes description

Add a new target option "Last valid answer" for categories and associated items.

This option pick the last answered question that is a valid target for this answer :
* Any dropdown question with the type "ITILCategory" for categories.
* Any glpiobject question with a type matching a valid asset itemtype for associated items.

### Checklist

### DIFF
--- a/inc/targetbase.class.php
+++ b/inc/targetbase.class.php
@@ -171,6 +171,7 @@ PluginFormcreatorConditionnableInterface
    const CATEGORY_RULE_NONE = 1;
    const CATEGORY_RULE_SPECIFIC = 2;
    const CATEGORY_RULE_ANSWER = 3;
+   const CATEGORY_RULE_LAST_ANSWER = 4;
 
    const LOCATION_RULE_NONE = 1;
    const LOCATION_RULE_SPECIFIC = 2;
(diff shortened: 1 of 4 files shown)


AdrienClairembault:
Waiting for your validation @btry before starting to work on the backport.

inc/targetbase.class.php
```diff
@@ -388,6 +390,55 @@ protected function setTargetCategory($data, $formanswer) {
          case self::CATEGORY_RULE_SPECIFIC:
             $category = $this->fields['category_question'];
             break;
+         case self::CATEGORY_RULE_LAST_ANSWER:
+            $category = null;
```
btry: a suggestion : move this line before the switch() {...} and drop the default case.

inc/targetbase.class.php
```diff
@@ -388,6 +390,55 @@ protected function setTargetCategory($data, $formanswer) {
...
+               ],
+               'ORDER' => [
+                  'row DESC',
+                  'col DESC',
+               ]
+            ]);
+
+            foreach ($answers as $answer) {
+               // Decode dropdown settings
+               $values = json_decode($answer['values']);
```
btry: in some cases the value may not be a JSON. I recently added a method to get the itemtype : PluginFormcreatorDropdownField::getSubItemtype()

To instanciate the field:

```php
$question = new PluginFormcreatorQuestion();
$question->getFromDB($questinId); // need to retrieve the id from the request a few lines earlier
$field = PluginFormcreatorFields::getFieldInstance($question->fields['fieldtype'], $question);
$itemtype = $field->getSubItemtype();
```

If you have an eligible question populated by the user then hidden by a condition (when the user fills other questions), I'm not sure that the value will be empty in DB then ignored by this code.

inc/targetbase.class.php
AdrienClairembault: Done.

inc/targetbase.class.php
AdrienClairembault: In this case I find it a bit non optimal to load two news objects from DB in order to access a value that I already have.
I've made a new static method in 5240c38 to avoid this issue, let me know if that's not ok for you.

AdrienClairembault:
> I think that js/scripts.js.php also needs an update
> 
>     * function plugin_formcreator_changeCategory
> 
>     * function plugin_formcreator_change_associate
> 
> 
> they show and hide the inputs for the various cases of category / item association settings

The default behavior of theses functions is to hide any extra on change then reload them if needed.
Since the new option doesn't use any extra this doesn't seem to be needed.

formcreator

inc/targetbase.class.php
btry: Ok for this commit.

btry:
To squash and merge.

PR #1688, opened 2020-02-18
Associated items were displayed incorrectly

TiagoDGomes:
inc/targetticket.class.php

161xEdSdEv


TiagoDGomes:
Oops:
`$options['items_id'][$row['itemtype']][$row['id']] = $row['items_id'];`

PR #1689, opened 2020-02-18
Fix image paste on multiple pages

AdrienClairembault:
Image paste is not working on multiple pages because `fileupload.js` is loaded before `tinymce.js`.
We need tinymce to be loaded first to set up our custom upload plugin:

#### fileupload.js
```js
/**
 * Plugin for tinyMce editor who intercept paste event
 * to check if a file upload can be proceeded
 * @param  {[Object]} editor TinyMCE editor
 */
if (typeof tinymce != 'undefined') {
    tinymce.PluginManager.add('glpi_upload_doc', function(editor) {
        editor.on('PastePreProcess', function(event) {


```

### Changes description

Force tinymce to be loaded first on the following pages :
* Form preview 
* Self service with formcreator interface
* Form in tech interface

### Checklist

### References

Possibly internal ref 18722 ?

### DIFF
--- a/setup.php
+++ b/setup.php
@@ -216,8 +216,16 @@ function plugin_init_formcreator() {
             ];
          }
 
-         if (strpos($_SERVER['REQUEST_URI'], 'plugins/formcreator/front/targetticket.form.php') !== false) {
-            Html::requireJs('tinymce');
+         $pages = [
+            "plugins/formcreator/front/targetticket.form.php",
+            "plugins/formcreator/front/formdisplay.php",
+            "plugins/formcreator/front/form.form.php"
+         ];
+         foreach ($pages as $page) {
+            if (strpos($_SERVER['REQUEST_URI'], $page) !== false) {
+               Html::requireJs('tinymce');
+               break;
+            }
          }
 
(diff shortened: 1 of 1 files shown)


btry:
This shoud be already fixed un support/2.9.0. See setup.php

AdrienClairembault:
The only reference to tinymce in setup.php is:
```php
if (strpos($_SERVER['REQUEST_URI'], 'plugins/formcreator/front/targetticket.form.php') !== false) {
    Html::requireJs('tinymce');
}
```

This does not cover the following pages;
* plugins/formcreator/front/formdisplay.php
* plugins/formcreator/front/form.form.php

Maybe i should add them to the condition instead ?

btry:
I checked and i see that the change I talked about is not here. I remember I updated the condition you just mentioned. I guess I did not merged this patch, it should be in a branch of my computer.

I prefer to put this change in setup, to reduce specificities in front/*.php to the strict minimum.

btry:
I updated the commit message of the PR and merged it locally (yesterday, there was a problem on GH preventing me to push commits). Consider this is merged.

https://github.com/pluginsGLPI/formcreator/commit/da30f86d6f9cb328a95579c97ef023367b37264a

PR #1695, opened 2020-03-03
textarea partially displayed

btry:
Change the way fields are hidden

If a textarea field is after a lot of other fields (and needs to scroll the webpage to show) and it is visible on condition, it may render partially and show only its toolbar. 

It seems hiding the textarea with display: none causes trouble with tinyMCE.


btry:
merged an updated version.

PR #1712, opened 2020-03-24
Rewrite backcreated code after post form to redirect to the ticket ch…

ddurieux:
…ange or the answer form

### DIFF
--- a/front/form.form.php
+++ b/front/form.form.php
@@ -142,21 +142,11 @@ if (isset($_POST['add'])) {
 
       // redirect to created item
       if ($_SESSION['glpibackcreated']) {
-         $item_ticket = new Item_Ticket;
-         $tickets = $item_ticket->find([
-            'itemtype' => PluginFormcreatorFormAnswer::class,
-            'items_id' => $formAnswer->getID(),
-         ]);
-
-         if (count($tickets) === 1) {
-            $current_ticket = array_pop($tickets);
-            $tickets_id = $current_ticket['tickets_id'];
-            Html::redirect(Ticket::getFormURLWithID($tickets_id));
+         if (count($formAnswer->targetList) == 1) {
+            $target = current($formAnswer->targetList);
+            Html::redirect($target->getFormURLWithID($target->getID()));
          }
-
-         if (count($tickets) > 1) {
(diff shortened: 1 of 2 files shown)


inc/formanswer.class.php
```diff
@@ -938,6 +940,10 @@ public function generateTarget() {
                $success = false;
                break;
             }
+            $this->targetList[] = [
+               'itemtype' => $generatedTarget->getType(),
+               'items_id' => $generatedTarget->getID()
+            ];
```
btry: Why not add the object instead of an array itemtype / items_id ? This way you could avoid the use of call_user_func_array in front/form.form.php

```suggestion
            $this->targetList[] = $generatedTarget;
```

front/form.form.php
```diff
@@ -142,21 +142,13 @@
...
-
-         if (count($tickets) === 1) {
-            $current_ticket = array_pop($tickets);
-            $tickets_id = $current_ticket['tickets_id'];
-            Html::redirect(Ticket::getFormURLWithID($tickets_id));
-         }
-
-         if (count($tickets) > 1) {
-            Html::redirect(PluginFormcreatorFormAnswer::getFormURLWithID($formAnswer->getID()));
+         print_r($formAnswer->targetList);
```
btry: I guess this is for debug purpose and should be removed

front/form.form.php
```diff
@@ -142,21 +142,13 @@
...
+         if (count($formAnswer->targetList) == 1) {
+            $target = current($formAnswer->targetList);
+            $url = call_user_func_array([$target['itemtype'], "getFormURLWithID"], [$target['items_id']]);
```
btry: This suggestion is related to the other suggestion below.

```suggestion
            $url = $target->getFormURLWithID($target->getID());
```

front/form.form.php
ddurieux: oh yes :p

inc/formanswer.class.php
ddurieux: seems a good idea

ddurieux: Fixed ;)

front/form.form.php
ddurieux: Fixed ;)

ddurieux:
Thanks ;)

PR #1713, opened 2020-03-24
FIx for asigned group from question answer

ddurieux:


btry:
Note : this fix was not backported from support/2.10. Now it's done.

PR #1722, opened 2020-04-07
fixUse font awesome to workaround pdf rendering problem

btry:
Fix pdf generation problem when using radio buttons


btry:
Replaced by #1732

PR #1757, opened 2020-05-11
Self service can close own ticket

btry:
In GLPI, a user can delete his ticket if he is in self service, the ticket is new and does not has any timeline item.

Formcreator shall allow this as well

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -207,6 +207,7 @@ function plugin_formcreator_addWhere($link, $nott, $itemtype, $ID, $val, $search
    switch ($table.".".$field) {
       case "glpi_plugin_formcreator_issues.status" :
          $tocheck = [];
+         /** @var CommonITILObject $item  */
          if ($item = getItemForItemtype($itemtype)) {
             switch ($val) {
                case 'all':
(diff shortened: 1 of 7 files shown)


hook.php
```diff
@@ -475,3 +483,24 @@ function plugin_formcreator_dynamicReport($params) {
...
+
+/**
+ * Hook for timeline_actions; display a new action for a CommonITILObject
+ * @see CommonITILObject
+ *
+ * @return void
+ */
+function plugin_formcreator_timelineActions($options) {
+   $item = $options['item'];
+   if (!$item->canDeleteItem()) {
```
btry, line 495: @orthagh : It looks like this method is a better choice than canRequesterUpdateItem() (by reading Ticket::showForm() )

orthagh: seems legit. I think there is in this function a repetition of code with canRequesterUpdateItem. It may need a factorization between the two on the core.

hook.php
```diff
@@ -475,3 +483,24 @@ function plugin_formcreator_dynamicReport($params) {
...
+      return;
+   }
+
+   if (!(isset($_SESSION['glpiactiveprofile']) &&
+       $_SESSION['glpiactiveprofile']['interface'] == 'helpdesk')) {
+      return;
+   }
+   echo "<li class='plugin_formcreator_cancel_my_ticket' onclick='".
+      "javascript:plugin_formcreator_cancelMyTicket(".$item->fields['id'].");'>"
+      ."<i class='fa'></i>".__('Put in trashbin')."</li>";
```
orthagh, line 505: I think "Cancel my ticket" is better label

PR #1760, opened 2020-05-12
Release 2.9.2

btry:


btry:
tests failures on GLPI 9.5. This is normal : the release flag is enabled and the release is officially not compatible with this version of GLPI.

Needs to find a way to avoid this false failure for next releases.

PR #1785, opened 2020-05-25
Feature/import dry run

btry:
### Changes description
Some users may encounteer import problems and the plugin leaves incomplete forms in the DB.

This PR adds an internal  dry-run mode to test an import and find errors without updating the database.


btry:
Should  be a better option to use SQL transaction.

PR #1787, opened 2020-05-26
Self service compatibility / improvements

AdrienClairembault:
### Changes description

- Not mandatory > Range check for checkboxes (if checkbox is not mandatory and empty, no need to check the range)
- Allow empty value for non mandatory date and datetime question (not sure if this was an API only bug but empty values were rejected as an invalid format)
- For dropdown question, the possible answers are included in the question data when retrieving it from the API.
This is needed as self-services users doesn't have access to these itemtypes (and can't be granted access, this is a limitation of self-services profiles).

Something like that might be also needed later for ObjectGLPI questions but for now they are still querying directly the API for the available items.

### Checklist

- [x] Docs have been added/updated (Not needed)

### DIFF
--- a/inc/fields/checkboxesfield.class.php
+++ b/inc/fields/checkboxesfield.class.php
@@ -210,6 +210,11 @@ class PluginFormcreatorCheckboxesField extends PluginFormcreatorField
          }
       }
 
+      // Not required and no answer -> valid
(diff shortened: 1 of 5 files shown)


AdrienClairembault:
Can we merge if I fix conflicts ?

btry:
Hi

Yes we can

PR #1800, opened 2020-06-08
fix(targettichet,targetchange): question tags not updated on duplication

### LINKED ISSUE #1797: Import formulaire - balise dans le titre non renumérotée
**Describe the bug**
Quand on importe un formulaire qui comporte une balise dans le titre, celle-ci n'est pas renumérotée comme dans la description.

**To Reproduce**
Créer un formulaire qui contient une balise (ou plusieurs) dans le titre.
Lors de l'import, celle(s)-ci ne sera(ont) pas renumérotées au contraire de celles de la description.

**Expected behavior**
Toutes les balises devraient être renumérotées lors de l'import.

**Additional context**
GLPI 9.4.6+FC 2.6.2

btry:
When duplicating a target the question tags are not updated

fix #1797


inc/targetchange.class.php
```diff
@@ -87,7 +87,7 @@ protected function getCategoryFilter() {
 
    protected function getTaggableFields() {
       return [
-         'name',
+         'target_name',
```
btry, line 90: target_name is the field in DB used to set the name of the taget to create.

PR #1823, opened 2020-06-17
Add full SLA/OLA support in target

AdrienClairembault:
Internal ref: 19538.

Add new options in targets to set service levels (SLA and OLA) or directly the expected dates (TTO, internal TTO, TTR and external TTR).
Full details regarding these new options can be found in the internal ref.

### DIFF
--- a/inc/fields/dropdownfield.class.php
+++ b/inc/fields/dropdownfield.class.php
@@ -62,6 +62,12 @@ class PluginFormcreatorDropdownField extends PluginFormcreatorField
       }
 
       $optgroup = Dropdown::getStandardDropdownItemTypes();
+
+      $optgroup[__('Service levels')] = [
+         SLA::getType() => __("SLA", "formcreator"),
+         OLA::getType() => __("OLA", "formcreator"),
+      ];
+
       $field = '<div id="dropdown_values_field">';
       $field .= Dropdown::showFromArray('dropdown_values', $optgroup, [
          'value'               => $itemtype,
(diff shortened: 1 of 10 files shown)


AdrienClairembault:
Unit tests seems to ignore 2_11 migration changes, not sure why.

btry:
I'll run then locally with your branch to find out the problem.

btry:
I think I found the problem @AdrienClairembault 

you added upgrade code, but you did not updated the SQL script for fresh installation. Because of that, there is a difference of schema between a fresh installation and an upgrade.

AdrienClairembault:
Can we merge if I drop the due_dates changes and keep the sla/ola part as discussed ?

btry:
Hi

Shall the plugin convert due dates into SLA ? This would allow us to drop due date.

AdrienClairembault:
I think we need to keep the due date (maybe deprecated it ?).
It can't always be replaced by SLA (e.g. DUE_DATE_RULE_ANSWER).

I'll put it back in if it's ok for you (the "new" internal_due_date, take_into_account_date and internal_take_into_account_date  are still removed).

AdrienClairembault:
Ready for review

btry:
Hi @AdrienClairembault 

I rebased and updated your PR to satisfy unit  tests.

Please have a review and if it is OK for you, merge it and ping me. I have a refactor PR to apply next (#1873)

AdrienClairembault:
ping @btry

PR #1835, opened 2020-06-25
Sql error when no entity restrict

btry:
When no entity restriction applies, the SQL query to get available categories is incorrect.


btry:
merged

PR #1841, opened 2020-06-29
Load tinymce

btry:
In service catalog, uploading a picture in a followup may fail.


btry:
conflict solved locally, and merged with 3d1ef1c9482fc1f2fb15765b54823fff5633c032

PR #1848, opened 2020-07-01
Fix plugin tag integration

AdrienClairembault:
### Changes description

Fix two bugs with "Tag" question type in 2.10.0 : 
- Method `parseObjectProperties` on parent `PluginFormcreatorDropdownField` was expecting an itemtype
- Available tags query was missing some values due to changes on the Tag plugin (0 -> empty string)


btry:
OK for me. Please merge and cherry-pick in branches 2.9 and develop

PR #1883, opened 2020-07-21
Fix mandatory field check on file field

### LINKED ISSUE #1882: [2.10.1] - Does not recognize attached file.
When a file is configured to be mandatory, the plugin does not recognize that it has been attached.

2020-07-20 11_13_00-Window

Thanks.

btry:
fix #1882


btry:
merged in support/2.10.0

PR #1896, opened 2020-07-27
Modal positioning

### LINKED ISSUE #1867: Modal in the top of the page GLPI 9.5 and FormCreator 2.10.0
Hi,

Using these version specified in the title, if we have a very large form, with several sections and questions and all properly configured by the rules, when clicking to add a section or question, the config modal is always displayed at the top of the page, being necessary to scroll up every time.

This is not a serious problem, but a nuisance, because if we are working with large forms, it is boring every time having to roll up until we reach the configuration mode.

What can I do to try make a correction?

btry:
some modals are not centered on screen

fix #1867

### DIFF
--- a/js/scripts.js.php
+++ b/js/scripts.js.php
@@ -80,7 +80,7 @@ $(function() {
       autoOpen: false,
       height: "auto",
       modal: true,
-      position: ['center', 50],
+      position: {my: 'center'},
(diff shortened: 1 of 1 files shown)


phamorais:
It's ok
Thank you @btry

PR #1911, opened 2020-08-07
Hide garbage conditions

btry:
In some cases (old versioins being upgraded) questions may have conditions in BD despite thei are always visible. since version 2.10.0 the conditions are visible. This path hides them.

The evaluation of conditions is not affected by the issue.

May consider a DB cleanup in the next version.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -457,6 +457,9 @@ input[type = radio]:checked + label.label-radio .check {
    font-style: italic;
    margin: 0 4px;
 }
+
+/* CONDITIONS */
+
 .div_show_condition_field, .div_show_condition_operator,
 .div_show_condition_value, .div_show_condition_logic,
 .div_show_condition_add, .div_show_condition_remove {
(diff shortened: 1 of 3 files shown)


css/styles.css
```diff
@@ -491,12 +494,12 @@ input[type = radio]:checked + label.label-radio .check {
    width: 12px;
    vertical-align: middle;
 }
-/* jQuery UI Combobox*/
-.pq-select-multiple {
-   display: inline-block;
-   width: 85% !important;
```
btry, line 497: obsolete CSS (long ago the lugin used PQ Select, from JQuery UI).

css/styles.css
```diff
@@ -491,12 +494,12 @@ input[type = radio]:checked + label.label-radio .check {
...
+tr[data-itemtype="PluginFormcreatorCondition"] .div_show_condition_logic {
+   visibility: hidden;
+}
+tr[data-itemtype="PluginFormcreatorCondition"] ~ tr[data-itemtype="PluginFormcreatorCondition"] .div_show_condition_logic {
+   visibility: visible;
```
btry, line 501: This hides the 1st boolean operator because it is meaningless, and got a report of user beiing disturbed about that.

js/scripts.js.php
```diff
@@ -834,13 +834,11 @@ function plugin_formcreator_addEmptyCondition(target) {
       data: data
    }).done(function (data) {
       $(target).parents('tr').after(data);
-      $('.plugin_formcreator_logicRow .div_show_condition_logic').first().hide();
```
btry, line 837: the CSS does the job better

js/scripts.js.php
```diff
@@ -834,13 +834,11 @@ function plugin_formcreator_addEmptyCondition(target) {
...
    });
 }
 
 function plugin_formcreator_removeNextCondition(target) {
    $(target).parents('tr').remove();
-   $('.plugin_formcreator_logicRow .div_show_condition_logic').first().hide();
```
btry, line 843: the CSS does the job better

PR #1917, opened 2020-08-14
Import progress bar

btry:
Add a progress bar for imports


inc/exportable.class.php
```diff
@@ -111,4 +112,42 @@ public function importChildrenObjects($item, $linker, $subItems, $input) {
...
+     */
+    public static function countChildren($input, $subItems = []) {
+        if (count($subItems) < 1) {
+            return 1;
+        }
+
+        $count = 0;
+        foreach($subItems as $key => $itemtypes) {
+           if (isset($input[$key])) {
+              // force array of itemetypes
```
AdrienClairembault, line 131:
```suggestion
              // force array of itemtypes
```

inc/exportableinterface.class.php
```diff
@@ -64,4 +64,11 @@ public static function import(PluginFormcreatorLinker $linker, $input = [], $con
     * @return boolean
     */
    public function deleteObsoleteItems(CommonDBTM $container, array $exclude);
+
+   /** get the count of inner objects to import
```
AdrienClairembault, line 68:
```suggestion
   /** 
     * Get the count of inner objects to import
```

inc/exportable.class.php
```diff
@@ -111,4 +112,42 @@ public function importChildrenObjects($item, $linker, $subItems, $input) {
             }
         }
     }
+
+    /**
+     * Count sub items
+     *
+     * @param array $input
+     * @param $subItems
+     * @return integer
```
AdrienClairembault, line 121: Not a big deal but PHP 7 only accept short terms (int, bool) in type hinting so it can be nice to be consistent with it.
Also recommended in [PSR 12](https://www.php-fig.org/psr/psr-12/#25-keywords-and-types).

inc/form.class.php
```diff
@@ -1979,6 +1986,18 @@ public static function import(PluginFormcreatorLinker $linker, $input = [], $con
...
 
+   public static function countItemsToImport($input) {
+      // Code similar to ImportChildrenObjects
+      $subItems = [
+         '_profiles'   => PluginFormcreatorForm_Profile::class,
+         '_sections'   => PluginFormcreatorSection::class,
+         '_conditions' => PluginFormcreatorCondition::class,
+         '_targets'    => (new self())->getTargetTypes(),
+         '_validators' => PluginFormcreatorForm_Validator::class,
+      ];
```
AdrienClairembault, line 1997: No ranges ? I think I used something like '_ranges' in the API.

inc/linker.class.php
```diff
@@ -120,15 +156,16 @@ public function linkPostponed() {
...
-            $newList = $postponedItemtypeList;
+            $newList = [];
             foreach ($postponedItemtypeList as $originalId => $postponedItem) {
                if ($itemtype::import($this, $postponedItem['input'], $postponedItem['relationId']) === false) {
                   $newList[$originalId] = $postponedItem;
                   $postponedAgainCount++;
                }
             }
+            $postponedItemtypeList = $newList;
          }
```
AdrienClairembault, line 170:
```suggestion
unset($postponedItemtypeList);
```
Can be a good idea to unset after using a reference in a foreach even if it's not needed in the current code.
This can avoid some nasty bugs if someone decide to add some code later and reuse this var.

See [php doc](https://www.php.net/manual/en/control-structures.foreach.php): `Reference of a $value and the last array element remain even after the foreach loop. It is recommended to destroy it by unset().`

inc/target_actor.class.php
```diff
@@ -181,6 +181,11 @@ public static function import(PluginFormcreatorLinker $linker, $input = [], $con
       return $itemId;
    }
 
+   public static function countItemsToImport($input)
```
AdrienClairembault, line 184: Wrong CS (I guess auto generated code by IDE ?).
Not the only case.

inc/linker.class.php
btry: +1, i'll add a commit for that

PR #1947, opened 2020-09-23
Fix footer

AdrienClairembault:
### Changes description

Before: 
image

After: 
image

### References

Internal ref: 20757.


btry (changes requested):
Hi

I disagree with your solution. I found that the HTML is broken. Please review my own PR. It looks smarter to me.

AdrienClairembault:
Replaced by #1949.

PR #1949, opened 2020-09-24
Broken footer

btry:
Internal ref 20757, alternative to #1947 

It lloks like HTML tag MAIN has been added (among others like HEADER).

There is a closing MAIN in the footer generated bi the plugin, because itreloes on GLPI core for this task. However, the header generated by the plugin was not updated to match changes in GLPI code. Header generated by the plugin cannot rely on the core.

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -1129,8 +1129,7 @@ a.plugin_formcreator_formTile_title {
    clear: both;
    position: fixed;
    bottom: 0;
-   width:  100%;
-   padding: 5px 0;
+   width:  calc(100% - 10px);
 }
 
 .plugin_formcreator_leftHeader#header .plugin_formcreator_leftMenu {
--- a/inc/wizard.class.php
+++ b/inc/wizard.class.php
@@ -163,6 +163,7 @@ class PluginFormcreatorWizard {
       self::showHeaderTopContent();
       echo '</div>'; //.formcreator_header_top
 
+      echo "<main role='main' id='page'>";
       echo '<div id="page" class="plugin_formcreator_page">';
 
       // call static function callcron() every 5min


AdrienClairembault (changes requested):
Ok for the fix.

However, I would keep some css changes from #1947.

Padding is mandatory IMO, the copyright is touching the right border of the screen.
Current:
image

With 10px padding: 
image

The z-index change is more subjective but I prefer the version with the sidebar on top.
Current: 
image

With side bar on top:
image

btry:
@AdrienClairembault : I updated the commit with some CSS tweaks.

css/styles.css
```diff
@@ -1129,15 +1129,14 @@ a.plugin_formcreator_formTile_title {
    clear: both;
    position: fixed;
    bottom: 0;
-   width:  100%;
-   padding: 5px 0;
+   width:  calc(100% - 10px);
```
AdrienClairembault, line 1132: I don't really like this solution as it depends on other values:
```css
#footer {
   font-size: 9px;
   color: #000;
   background-color: #E2E1DE;
   padding: 6px;
   height: 18px;
   padding: 5px;  <-----------------------------------

   > table {
      width: 100%;
   }
}
```

If someone change this padding, the calc will not work as expected as it need 100% - (padding-left + padding right).

If you really want to do it this way, maybe use CSS vars ?

css/styles.css
```diff
@@ -1129,15 +1129,14 @@ a.plugin_formcreator_formTile_title {
...
 }
 
 .plugin_formcreator_leftHeader#header .plugin_formcreator_leftMenu {
    width: 300px;
    padding-left: 0; /* Overrides CSS from GLPi */
    padding-top: 20px;
-   height: 100%;
+   height: calc(100% - 20px - 65px - 28px); /* - height of top area - padding - footer */
```
AdrienClairembault, line 1139: Same as other comment, depends on other values.

Also this doesn't change anything visually for me (the height of the menu is reduced to not overlap with the footer but it isn't noticeable since the footer was displayed on top of it).

btry:
I would also prefer CSS vars. I considered that approximatively  a year ago, but I thought this was too early because of browser compatibility. 

Moreover, the plugin is still compatible with GLPI 9.4 then we are stuck to the lowest browser requirements of this version. I need to check those lowest versions to determine the ability to safely use CSS vars.

If not allowed right now, I suggest to stick to the current CSS, then switch to CSS vars as soon the plugin requires a recent enough GLPI. To not forget this CSS update, we may fill an issue.

css/styles.css
btry: I remove it.

AdrienClairembault (approved):

PR #1958, opened 2020-10-05
Set template ticket from category ticket

MLOcc:
### Changes description

Change of save function in PluginFormcreatorTargetTicket class :
When the designer of a form doesn't set a template ticket value in target ticket 
image, 
the template ticket used to generate the ticket results from templates tickets set in the ticket category according to the type of ticket.
image

### References

#1951

### DIFF
--- a/css/styles.css
+++ b/css/styles.css
@@ -1129,8 +1129,7 @@ a.plugin_formcreator_formTile_title {
    clear: both;
    position: fixed;
    bottom: 0;
-   width:  100%;
-   padding: 5px 0;
+   width:  calc(100% - 10px);
 }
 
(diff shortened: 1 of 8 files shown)


btry:
Hi

I reworked your PR to fix code style, and to base it on top of develop branch.

I cannot merge your contribution in support/2.10.0 because it introduces a critical change og behavior if a target ticket has no template set and the category of the generated ticket has templates set. Currently, users may want to ignore the templates set in ITIL categories.

The feature needs a new option to allow that (a new checkbox labelled "Ignore templates in categories" or similar. By the way, this checkbox needs a new column in DB, This confirms the merge in  the next development iteration (DB tables should not change in bugfix releases).

btry:
Obsoleted by my reworked version. Closing.

PR #1965, opened 2020-10-09
Provide hook for external plugins

btry:
Give a hook to other plugins


btry:
Implemented with: 91926b29b04049f1ca3d2020b5faaa657637272e

obsolete

PR #1983, opened 2020-10-21
Fix update and purge hooks

AdrienClairembault:
### Changes description

Bug introduced in 4819b268824f45da3f97ca74ff7bcc858dc67c2c.
`item_update` and  `item_purge` hooks were declared two times.
Fixed by merging the two hooks.


AdrienClairembault:
I think we are going to need a 2.10.4 release for this fix.

AdrienClairembault:
And the update should re-sync all issues.

btry (approved):

PR #1987, opened 2020-10-26
Bugfix/1960 reservation search

### LINKED ISSUE #1960: Asset booking search does not work in the simplified interface
Good Morning,

I noticed that when I'm in the simplified interface, asset search doesn't work, because it returns all assets including those that are reserved and shouldn't be showing up.
Do you know if you can fix this?

GLPI 9.5.1
FormCreator 2.10.2

btry:
fix #1960

### DIFF
--- a/front/reservationitem.php
+++ b/front/reservationitem.php
@@ -44,6 +44,11 @@ PluginFormcreatorWizard::header(__('Service catalog', 'formcreator'));
 $res = new ReservationItem();
 $res->display($_GET);
 
+if (isset($_GET['reset'])) {
+   unset($_SESSION['plugin_formcreator']['redirected']);
+}
+$_POST = $_SESSION['plugin_formcreator']['redirected']['POST'] ?? [];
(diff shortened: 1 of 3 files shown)


front/reservationitem.php
```diff
@@ -44,6 +44,8 @@
 $res = new ReservationItem();
 $res->display($_GET);
 
+$_POST = $_SESSION['plugin_formcreator']['redirected']['POST'] ?? [];
```
AdrienClairembault, line 47: Is it possible to go to this page without being redirected ?
In this case $_POST would be erased.

btry: Yes this is possible when the user clicks on the menu entry to view the reservations. In this case there is no search to recover. 

I amended my commit to reset the session in this case, to avoid to recover an obsolete search. This is hacky and in some cases this reset will not work.

AdrienClairembault (approved):

PR #1993, opened 2020-10-28
Update dependencies

btry:
Needed to solve switch to composer 2 on travis-ci


btry:
travis dropped, and updates donne while switching to GHA

PR #1998, opened 2020-11-02
Use GLPI's background  color

btry:
better compatibility with the GLPI's color palette


btry:
obsoleted : due to switch to scss in develop branch.

PR #2025, opened 2020-11-24
Translatable form

btry:
Make forms usable with several languages.

should supprezss need to duplicate and maintain several copies of the same form because users may use different  languages

UX still  needs improvement


btry:
replaces by #2064

PR #2029, opened 2020-11-26
Do not load non existing JS lib

btry:


ladenree76000:
using glpi9.4.6 and Formcreator 2.10.4 :

proposed fix did not solve

setting lines 262 & 263 as comment did solve

something is probably wrong with condition in fix :  " if (version_compare(GLPI_VERSION, '9.4') > 0) "

btry:
Well.. i reversed the comparison operator. It should be <= , not > .

btry:
@ladenree76000 : I updated the fix. This should work now.

PR #2031, opened 2020-12-01
[9.4 BC BREAK] Inline imagein default value converted to document

btry:
- Fix inline image handling in textareas, uploads are converted into documents
- default value inline image converted into document in default value of textarea

this solve a missing image problem in email notification of tickets generated by formcreator, when viewed from a webmail such Gmail.

Should be merged ONLY in develop branch, when GLPI 9.4 compatibility is dropped.


btry:
cherry-picked in develop. Won't merge in support/2.10.0 because of possible regression with GLPI 9.4.

Kept here if needed in 2.10.0 + GLPI 9.5

PR #2041, opened 2020-12-11
Issue status computation

btry:
Fix an error and support for multiple validations


btry:
Obsolete; done in 2.11.

PR #2056, opened 2020-12-21
Captcha for anonymous forms

btry:
Spam prevention with captcha.

ReCaptcha not used because if an online dependency, and no feature yet to enabie / disable the feature on on a lower granularity

image


btry:
New feature : rebased on develop branch and merged with an extra enhancement. See #2063

PR #2086, opened 2021-01-18
Missing rows in issues after repopulate

btry:
see #1948 

Patch for version support/2.11.0 and develop branches


btry:
Hi

I found some similar cases where I could directly investigate in  database. The regression since 2.10.2 is confirmed, the above fixes seems to work properly. 

Having no feedback and having patches which did the job on instances where I can investigate, I think I can close.

A fix was added in the branch support/2.10.0, but will not be released, as the focus is now on 2.11.0 (whci contains the similar fix, obviously).

PR #2138, opened 2021-02-18
Don't show tabs for KB item in service catalog

btry:
### before

image

### after 

image


btry:
closed due to bad branch name.Replaced by #2171

PR #2146, opened 2021-02-23
Simplify kb display

btry:
Remove KB tabs but keep documents


btry:
replaced by: 514905d25fc09dd70ceb455208dfce68d4006acb

PR #2160, opened 2021-03-04
Fix trailing commas in function

AdrienClairembault:
### Changes description

Trailing commas in function call aren't supported by all browsers:
image

### References

Internal ref: 21303.


AdrienClairembault:
Default parameters aren't supported too:
image

PR #2164, opened 2021-03-08
Dropdown recursivity

btry:
Disable non conform recutrsivity in dropdown, solve a IDOR issue


btry:
should be related to https://github.com/glpi-project/glpi/issues/8832

PR #2173, opened 2021-03-15
Redirect to next form

xacobofg:
### Changes description

Adds the possibility to redirect to another form when a form is completed.

### DIFF
--- a/front/form.form.php
+++ b/front/form.form.php
@@ -142,6 +142,10 @@ if (isset($_POST['add'])) {
    }
    $form->increaseUsageCount();
 
+   if ($form->fields['plugin_formcreator_forms_id'] > 0) {
+      Html::redirect('formdisplay.php?id=' . $form->fields['plugin_formcreator_forms_id']);
+   }
+
    if ($_SESSION['glpiname'] == 'formcreator_temp_user') {
       // Form was saved by an annymous user
       unset($_SESSION['glpiname']);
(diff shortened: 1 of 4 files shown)


xacobofg:
Hi @btry 
I changed the order of the fields.
What do I have to modify in the test file?

btry:
Hi

Check the test execution log of your PR (latest run is here : https://github.com/pluginsGLPI/formcreator/pull/2173/checks?check_run_id=2119989310)

Exporting a form is broken because you aded a new field. Tests need to be updated to to recognize this new field

image

Also as the new field is a foreign key to a form, you need to add logic to
- convert a form ID into an UUID on export
- convert an UUID into a form ID on import.

See the export and import methods of PluginFormcreatorForm to apply those changes. This feature is implemented in a rather compicated way. If you don't understand well how it works, I'll handle this, but it may take time  before I find enough time for tis task. The target version for your feature is 2.12.0.

xacobofg:
Hi @btry 
I modified the tests but it failed again.
Can you tell me what needs to be modified?

btry:
I need to push this contribution to a later version : a few things to solve vefore to merge
- import of forms
- deletion of forms
- infinite loops across linked forms

btry:
Hi

for now there is an important problem with this feature. some users will use it a lot, and there is a risk of infinite loop. For example 3 forms A,B and C with A next B next C.

This is the main reason i did not merge this feature. It is incomplete for now and I don't had enough time to add code to avoid such loop.

derry43:
Infinite loops could be prevented by maintaining a list of form IDs in the chain and passing this to each target form. This list could be checked to see whether the next target form has already been a target in the chain and the chain stopped with an error.

PR #2229, opened 2021-04-27
Ensure unique order for duplicate

### LINKED ISSUE #2228: [2.11.2] Using duplicate icon on section set the same order value to the new section
**Describe the bug**
When using duplicate icon on a section, the new section have the same order value than the duplicated one in glpi_plugin_formcreator_sections.
Sorting sections become impossible or have bad behavior.

**To Reproduce**
Add a section named "section 1"
image
Use duplicate icon and rename new section "section 2"
image
Save
image
It's not possible to set "section 2" before "section 1"

Add a new section named "section 3"
image
Use duplicate icon on "section 3" and rename new section "section 4"
image
Save

Set "section 1" order after "section 2" and before "section 3" using arrow icon
image

Save
image
image
It displays sections 2-3-1-4 instead of 2-1-3-4

**Expected behavior**
When using duplicate icon on section A, the new section B must have set "order of B"="order of A"+1 and all the sections C having their order > "order of A" have to be updated with "order of C"="order of C"+1

Best regards

btry:
fix #2228

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -401,12 +401,10 @@ PluginFormcreatorConditionnableInterface
       if (!isset($input['width'])) {
          $input['width'] = PluginFormcreatorSection::COLUMNS - $input['col'];
       }
-      $sectionFk = PluginFormcreatorSection::getForeignKeyField();
       // Get next row
       if ($this->useAutomaticOrdering) {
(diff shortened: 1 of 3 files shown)


btry:
merged with improvements : 9db4229a99eb401141185287c0e366e95e18d382

PR #2297, opened 2021-07-07
Added actor type: "Form authors supervisor"

aagz:
User from 'Responsible' field

### DIFF
--- a/inc/abstracttarget.class.php
+++ b/inc/abstracttarget.class.php
@@ -701,6 +701,16 @@ PluginFormcreatorTranslatableInterface
 
                $userIds = [$object->fields[$groupFk]];
                break;
+
+            case PluginFormcreatorTarget_Actor::ACTOR_TYPE_AUTHORS_SUPERVISOR:
+               $requester_id = $formanswer->fields['requester_id'];
+
+               $user = new User;
+               $user = User::getById($requester_id);
+               if (is_object($user)) {
+                  $userIds = [$user->fields['users_id_supervisor']];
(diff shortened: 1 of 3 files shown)


inc/abstracttarget.class.php
```diff
@@ -696,6 +696,17 @@ protected function prepareActors(PluginFormcreatorForm $form, PluginFormcreatorF
...
+
+            case PluginFormcreatorTarget_Actor::ACTOR_TYPE_AUTHORS_SUPERVISOR:
+               $requester_id = $formanswer->fields['requester_id'];
+
+               $user = new User;
+               $users = $user->find(['id' => $requester_id]);
+               if (!empty($users)) {
+                  $userIds = [$users[$requester_id]['users_id_supervisor']];
+               }
+               // $userIds = [$formanswer->fields['requester_id']];
```
btry, line 708: $user = User::getById($requester_id);
```suggestion
               $user = User::getById($requester_id);
               if ($user !== false) {
                  $userIds = [$user->fields['users_id_supervisor']];
               }
```

btry (commented):
Looks good, but a part of code could be more readable. See my suggestion; I think it should work.

inc/abstracttarget.class.php
aagz: Thanks, it works.

aagz (commented):
Thank you.

btry:
I rebase your contribution onto 2.12.0 because this is a small feature.

PR #2336, opened 2021-08-03
Redirect technicians to correct URL

AdrienClairembault:
### Changes description

It is common for requesters to reference other tickets in their tickets description or followups.
These references will be links like this one: `http://localhost/g95x/plugins/formcreator/front/issue.form.php?id=282`.

This is an issue because this link cannot be used by a technician: they are given an error and have no clue how to get the real linked ticket.

To fix this, I've added a redirect for any user coming from the 'central' interface.

### References

!22442

### DIFF
--- a/front/issue.form.php
+++ b/front/issue.form.php
@@ -37,6 +37,15 @@ if (!(new Plugin())->isActivated('formcreator')) {
    Html::displayNotFoundError();
 }
 
+// Accessing an issue from a tech profile, redirect to ticket page
+if (isset($_REQUEST['id']) && Session::getCurrentInterface() == 'central') {
+   /** @var PluginFormcreatorIssue $issue */
+   $issue = PluginFormcreatorIssue::getById((int) $_REQUEST['id']);
+   $id = $issue->fields['items_id'];
+   $itemtype = strtolower($issue->fields['itemtype']);
+   Html::redirect($CFG_GLPI['root_doc'] . "/front/$itemtype.form.php?id=$id");
+}
+
 // Show issue only if service catalog is enabled
 if (!plugin_formcreator_replaceHelpdesk()) {
    Html::redirect($CFG_GLPI['root_doc']."/front/helpdesk.public.php");


front/issue.form.php
```diff
@@ -48,14 +48,22 @@
    $issue->saveAnswers($_POST);
    Html::back();
 } else {
+   $issue->getFromDB((int) $_REQUEST['id']);
+
+   // Accessing a form from a tech profile, redirect to ticket page if possible
+   if (Session::getCurrentInterface() == 'central') {
+      $id = $issue->fields['original_id'];
```
btry, line 55: **original_id** and **sub_itemtype** were renamed respectively to **items_id** and **itemtype**

Also, for maintenance readability, I think this redirection should be located in setup.php, after the enlightened if() statement.

image

btry:
risk of conflict with : https://github.com/pluginsGLPI/formcreator/pull/2321

front/issue.form.php
AdrienClairembault: > original_id and sub_itemtype were renamed respectively to items_id and itemtype

Good catch, I tested on the previous version and assumed it didn't changed in 2.12... should have tested on this version too :)
Fixed fa7ee21b.

For the redirection I can change it if you want but it seemed simpler for me to have it in the file directly targeted by the link.
-> Make it easier for a developer to finds it as you can directly check the file you were trying to access (issue.form.php) instead of searching for another unrelated file.

It is different that the others redirect done in `setup.php` that target core files and thus can't be done directly in them since we are in a plugin.

btry: OK, this is not a big concern anyway, but let's wait for the merge of #2321 to rebase this one and see if there is a conflict.

btry:
Like I expected, thre is conflict with an other PR.

I believe this diff is equivalent to your patch. It implements the redirection in setup.php, and is based in the current HEAD of support/2.12.0

```diff
diff --git a/setup.php b/setup.php
index 5873190c..408b51df 100644
--- a/setup.php
+++ b/setup.php
@@ -247,6 +247,16 @@ function plugin_init_formcreator() {
                   break;
                }
             }
+         } else if (Session::getCurrentInterface() == 'central')  {
+            if (strpos($_SERVER['REQUEST_URI'], FORMCREATOR_ROOTDOC . '/front/issue.php') !== false) {
+               /** @var PluginFormcreatorIssue $issue */
+               $issue = PluginFormcreatorIssue::getById($_GET['id']);
+               if ($ssue === false) {
+                  Html::displayNotFoundError();
+               }
+               $itemtype = strtolower($issue->fields['itemtype']);
+               Html::redirect($itemtype::getFormURLWithID($issue->fields['items_id']));
+            }
          }
 
          // Massive Action definition
```

AdrienClairembault:
Conflict fixed 👍

PR #2393, opened 2021-09-10
Showfields access check

btry:
check access rights before sending fields visibility information


ajax/showfields.php
```diff
@@ -31,6 +31,39 @@
 
 include ('../../../inc/includes.php');
 
+$criteria = [
```
orthagh, line 34: the below code is good, but i think it could be better in a dedicated function (can be at least unit tested)

btry: I agree : this is a copy of code found in front/formdisplay.php with modifications dedicated to use with AJAX.

As I wand to avoid a regression, I would defer factorisation in the develop branch.

orthagh (approved):

btry:
only the last commid is related to the title of the PR, others are here for unknown reason. To avoid mess, the commit was cherry-picked. 

Closing.

PR #2466, opened 2021-11-18
Add validation setting to targets

AdrienClairembault:
### Changes description

Add new "Validation" setting for targets:
image

Possible values :
- User:  
image
- Group:  
image
- From question:
image

### References

!22598


inc/abstracttarget.class.php
```diff
@@ -1501,6 +1512,79 @@ protected function showLocationSettings($rand) {
...
+      // Hidden secondary values, displayed according to the user main choice
+      echo '<td width="25%">';
+
+      // VALIDATION_RULE_SPECIFIC_USER_OR_GROUP
+      echo '<div id="validation_specific" style="display: none">';
+      $validation_dropdown_params = [
+         'name' => 'validation_specific'
+      ];
+      $validation_data = json_decode($this->fields['validation_question'], true);
+      if (!is_null($validation_data) && isset($validation_data['type'])) {
```
btry, line 1551: when ```isset($validation_data['type'])``` is true, ```is_null($validation_data) ``` should be always false,  then this is redundant

```suggestion
      if (isset($validation_data['type'])) {
```

inc/abstracttarget.class.php
```diff
@@ -1501,6 +1512,79 @@ protected function showLocationSettings($rand) {
...
+      echo '<td width="15%">' . __('Validation') . '</td>';
+
+      // Possible values
+      echo '<td width="45%">';
+      Dropdown::showFromArray('validation_rule', static::getEnumValidationRule(), [
+         'value'     => $this->fields['validation_rule'],
+         'on_change' => "plugin_formcreator_change_validation($rand)",
+         'rand'      => $rand
+      ]);
+      echo Html::scriptBlock("plugin_formcreator_change_validation($rand)");
```
btry, line 1528: I assume this JS function manipulates display attribut on subsequent DIV tags. To speedup clisnt side rendering, I think it would be better to set display attribute properly server side.

inc/abstracttarget.class.php
```diff
@@ -1501,6 +1512,79 @@ protected function showLocationSettings($rand) {
...
+               JSON_OBJECT_AS_ARRAY
+            );
+            if (($decoded_values['itemtype'] ?? "") === 'User') {
+               $user_questions[$id] = $form_question->fields['name'];
+            }
+         }
+      }
+      Dropdown::showFromArray('_validation_from_user_question', $user_questions, [
+         'value' => $this->fields['validation_question'],
+      ]);
```
btry, line 1581: PluginFormcreatorQueston::dropdownForForm() should help to replace this big snippet of code.

The following should work, worth to test it and adjust if necessary.

```suggestion
      PluginFormcreatorQuestion::dropdownForForm(
         $this->getForm()->getID(),
         [
            new QueryExpression("`fieldtype` = 'actor' OR `fieldtype` = 'glpiselect' AND `itemtype`='User'"),
         ],
         'value' => $this->fields['validation_question'],
      );
```

install/mysql/plugin_formcreator_empty.sql
```diff
@@ -179,6 +179,8 @@ CREATE TABLE IF NOT EXISTS `glpi_plugin_formcreator_targetchanges` (
   `tag_specifics`              varchar(255) NOT NULL DEFAULT '',
   `category_rule`              int(11) NOT NULL DEFAULT '1',
   `category_question`          int(11) NOT NULL DEFAULT '0',
+  `validation_rule`            int(11) NOT NULL DEFAULT '1',
```
btry, line 182: I think the columns should be renamed : we don't know if the validation is related to form answer or ticket. In the future, we may have headache to maintain and debug.

- ticket_validation_rule
- ticket_validation_question

inc/abstracttarget.class.php
```diff
@@ -195,6 +195,10 @@ abstract protected function getTargetTemplate(array $data): int;
    const LOCATION_RULE_SPECIFIC = 2;
    const LOCATION_RULE_ANSWER = 3;
 
+   const VALIDATION_RULE_NONE = 1;
+   const VALIDATION_RULE_SPECIFIC_USER_OR_GROUP = 2;
+   const VALIDATION_RULE_ANSWER_USER = 3;
```
btry, line 200: I think the columns should be renamed : we don't know if the validation is related to form answer or ticket. In the future, we may have headache to maintain and debug.

- TICKET_VALIDATION_RULE_NONE
- TICKET_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP
- TICKET_VALIDATION_RULE_ANSWER_USER

btry (changes requested):

install/mysql/plugin_formcreator_empty.sql
AdrienClairembault: Would "comonitil_validation_rule" be ok instead ? Since it applies to both change and tickets.

btry: good point. Ok for me

inc/abstracttarget.class.php
AdrienClairembault: `itemtype` is a JSON property of the `values` field.
I don't think it can be handled with a simple SQL criteria (`fieldtype` = 'glpiselect' AND `itemtype`='User').

btry: This no longer true. I changed this recently to solve similar need elsewhere in the plugin.

  https://github.com/pluginsGLPI/formcreator/blob/59ea8b79f00e2ca9017754489f70a08f3406eed6/inc/field/dropdownfield.class.php#L891 

https://github.com/pluginsGLPI/formcreator/blob/59ea8b79f00e2ca9017754489f70a08f3406eed6/install/upgrade_to_2.13.php#L69

AdrienClairembault: Ok good to know, done.

inc/abstracttarget.class.php
```diff
@@ -1634,6 +1759,24 @@ public function prepareInputForUpdate($input) {
          $input['uuid'] = plugin_formcreator_getUuid();
       }
 
+      switch ($input['commonitil_validation_rule']) {
```
btry, line 1762: There is a failure in unit tests here : This code makes commonitil_validation_rule mandatory. As some users begin to use the plugin via API, it would be better to check if commonitil_validation_rule and commonitil_validation_question are set before processing them. This allows an API client to not mention them if the defaults are OK.

inc/abstracttarget.class.php
```diff
@@ -1634,6 +1759,24 @@ public function prepareInputForUpdate($input) {
...
+         default:
+         case self::COMMONITIL_VALIDATION_RULE_NONE:
+            $input['commonitil_alidation_question'] = '0';
```
btry, line 1765:
```suggestion
            $input['commonitil_validation_question'] = '0';
```

btry (changes requested):

btry:
#2496 merged : this is a rebase of this PR.

PR #2482, opened 2021-11-30
Drop ldap restrition

btry:
internal ref 22945 23259

### DIFF
--- a/inc/field/ldapselectfield.class.php
+++ b/inc/field/ldapselectfield.class.php
@@ -53,14 +53,9 @@ class LdapselectField extends SelectField
       if ($ldap_values === null) {
          $ldap_values = [];
       }
-      $current_entity = $_SESSION['glpiactive_entity'];
-      $auth_ldap_condition = '';
-      if ($current_entity != 0) {
-         $auth_ldap_condition = "glpi_authldaps.id = (select glpi_entities.authldaps_id from glpi_entities where id=${current_entity})";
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
The changes were introduced by https://github.com/pluginsGLPI/formcreator/pull/1437.
Are the discussion in this PR no longer valid ? 
Wouldn't this change allow you to see an LDAP from another entity that you do not have access to ?.

btry:
I see several problems with this implemetnation

- access to all LDAP when the user is in the root entity. The root entity does not grands more privileges than others.

- LDAPs are not linked to an entity, then entity restriction is not applicable.

An entity may be linked to a LDAP to search for users, this type of relation should not be used to restrict the available directoeirs when editing a question (see below)

image

To edit forms, the user needs to have UPDATE right on entities. This means that the user also have rights to change the LDAP associated to the current entity (and maybe others). 

Maybe LDAP questions should be restricted to people able to edit general config of GLPI. See https://github.com/glpi-project/glpi/blob/9.5/bugfixes/front/authldap.form.php#L35

AdrienClairembault:
> LDAPs are not linked to an entity, then entity restriction is not applicable.
Ok, seems fine to DROP this restriction then.

It does feel bad not being able to filter LDAP on some criteria but if nothing exist we can't do much.

AdrienClairembault (approved):

PR #2494, opened 2021-12-10
Drop upgrade support from very old versions (< 2.5.0)

btry:
Lowest supported version for upgrade is now 2.5.0

image

image


orthagh (approved):

AdrienClairembault (approved):

inc/field/ldapselectfield.class.php
```diff
@@ -49,7 +49,7 @@ public function getDesignSpecializationField(): array {
       $label .= _n('LDAP directory', 'LDAP directories', 1);
       $label .= '</label>';
 
-      $ldap_values = json_decode(plugin_formcreator_decode($this->question->fields['values']), JSON_OBJECT_AS_ARRAY);
```
btry, line 52: plugin_formcreator_encode() is not used in method PrepareQuestionInputForSave() then we should be able to remove plugin_formcreator_decode()

inc/field/ldapselectfield.class.php
```diff
@@ -120,7 +120,7 @@ public function getAvailableValues() {
          return [];
       }
 
-      $ldap_values   = json_decode(plugin_formcreator_decode($this->question->fields['values']));
```
btry, line 123: plugin_formcreator_encode() is not used in method PrepareQuestionInputForSave() then we should be able to remove plugin_formcreator_decode()

btry:
review request again as there are residual calls to plugin_formcreator_encode() in ldap field.

PR #2495, opened 2021-12-13
Adapt to new GLPI's autoload

btry:
### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/abstractquestionparameter.class.php
+++ b/inc/abstractquestionparameter.class.php
@@ -65,7 +65,7 @@ PluginFormcreatorTranslatableInterface
     *                - fieldName: name of the HTML input tag
     *                - label    : label for the parameter
     */
-   public function __construct(PluginFormcreatorFieldInterface $field, array $options) {
+   public function setField(PluginFormcreatorFieldInterface $field, array $options) {
       $fieldType = $field->getFieldTypeName();
       $fieldName = $options['fieldName'];
       $this->domId = $this->domId . "_{$fieldType}_{$fieldName}";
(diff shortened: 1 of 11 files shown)


inc/abstractquestionparameter.class.php
```diff
@@ -65,7 +65,7 @@ abstract class PluginFormcreatorAbstractQuestionParameter
     *                - fieldName: name of the HTML input tag
     *                - label    : label for the parameter
     */
-   public function __construct(PluginFormcreatorFieldInterface $field, array $options) {
+   public function setField(PluginFormcreatorFieldInterface $field, array $options) {
```
btry, line 68: The signature of this method is diferent from the signature in the parent classes. PHP now complains, and this needs a fix.

inc/question.class.php
```diff
@@ -442,9 +442,8 @@ public function prepareInputForAdd($input) {
       // handle description field and its inline pictures
       if (isset($input['_description'])) {
          foreach ($input['_description'] as $id => $filename) {
-            // TODO :replace PluginFormcreatorCommon::getDuplicateOf by Document::getDuplicateOf
-            // when is merged https://github.com/glpi-project/glpi/pull/9335
```
btry, line 446: this PR is merged in GLPI 10; and GLPI 9.5 support is dropped.

tests/src/CommonTestCase.php
```diff
@@ -45,8 +45,6 @@ protected function setupGLPIFramework() {
 
       $DB = new DB();
 
-      include_once (GLPI_ROOT . "/inc/timer.class.php");
```
btry, line 48: breaks tests, and should be useless as the autoloader is already enabled. I don't remember why this was added.

AdrienClairembault (approved):

PR #2497, opened 2021-12-13
Add form visibility setting

AdrienClairembault:
### Changes description

Add a new "Visible" setting to forms.
If this setting is disabled, the form wont be visible in the wizard but can still be accessed directly through its URL.

### Checklist

### References

!22487

Closes #N/A


AdrienClairembault:
Correct me if I'm wrong but tests failure doesn't seem related to these changes.

btry (changes requested):
You need to update export and import unit tests. There is also a fixture form containing al types of questions. This is an export JSON, then it must be updated with the new field.

The code itself is ok.

btry:
> Correct me if I'm wrong but tests failure doesn't seem related to these changes

Ys, the failures are not related to your code. I'm searching the solution as it occurs in other branches.That said if the tests to export / import a form don't complain with your new code, this is unusual, this is quite strange.

AdrienClairembault:
My bad I didn't see the export part in the CI results, only this output which I though was unrelated:
image

PR #2500, opened 2021-12-15
Show ticket in service catalog

btry:
Some methods are deleted in GLPI 10, breaking the plugin

Clean up scripts as GLPI 9.5 is no longer supported

### DIFF
--- a/front/issue.form.php
+++ b/front/issue.form.php
@@ -53,10 +53,6 @@ if (!plugin_formcreator_replaceHelpdesk()) {
    Html::redirect($CFG_GLPI['root_doc']."/front/helpdesk.public.php");
 }
 
-// force layout of glpi
-PluginFormcreatorCommon::saveLayout();
-$_SESSION['glpilayout'] = "lefttab";
-
 /** @var PluginFormcreatorIssue $issue */
 $issue = PluginFormcreatorIssue::getById((int) $_REQUEST['id']);
 
@@ -91,7 +87,6 @@ if ($issue) {
    } else {
       Html::header($header);
    }
-   PluginFormcreatorCommon::restoreLayout();
    Html::displayNotFoundError();
 }
 
@@ -100,5 +95,3 @@ if (Session::getCurrentInterface() == "helpdesk") {
 } else {
    Html::footer();
 }
-
-PluginFormcreatorCommon::restoreLayout();
--- a/front/issue.form.php
+++ b/front/issue.form.php
@@ -48,11 +48,6 @@ if (isset($_REQUEST['id']) && Session::getCurrentInterface() == 'central') {
    Html::redirect($itemtype::getFormURLWithID($id));
 }
 
-// Show issue only if service catalog is enabled
-if (!plugin_formcreator_replaceHelpdesk()) {
-   Html::redirect($CFG_GLPI['root_doc']."/front/helpdesk.public.php");
-}
-
 /** @var PluginFormcreatorIssue $issue */
 $issue = PluginFormcreatorIssue::getById((int) $_REQUEST['id']);
 
--- a/inc/common.class.php
+++ b/inc/common.class.php
@@ -598,14 +598,6 @@ JAVASCRIPT;
       return $isValid;
    }
 
-   public static function saveLayout() {
-      $_SESSION['plugin_formcreator']['layout_backup'] =  $_SESSION['glpilayout'];
-   }
-
-   public static function restoreLayout() {
-      $_SESSION['glpilayout'] = $_SESSION['plugin_formcreator']['layout_backup'] ?? $_SESSION['glpilayout'];
-   }
-
    /**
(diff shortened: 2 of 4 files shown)


AdrienClairembault (commented):
I only checked the simplified interface for now.
-> There is a lot more data displayed than in the 9.5, is this on purpose ?

image

We can see assigned tech and much more.
In 9.5 you could only see the timeline.

btry:
if service catalog is extended we call $item-showTabsContent() then no change to do here

If we use simplified service catalog we want to show the timeline. Up to GLPI 9.5 timeline was an additional tab. 

In GLPI 10, the timeline belongs to the main tab, then I think it is nice to do as I propose.
image

If I compare the UI in simplified interface and the result of this PR it appears we have the same ticket's form.
image

image

AdrienClairembault:
Is there no problems with exposing this additional data ?
Maybe some admins use the simplified mode to add these informations ?
What do you think @orthagh ?

Also I think the timeline side is missing it's container, visuals are different in your two screenshots.

AdrienClairembault (changes requested):
A few minor visual issues:

Missing margin:
image

Core for comparison:
image


Status is not vertically centered (possibly a core issue on read-only dropdown) :
image

I have access to the "pending" toggle, should not be possible for a self service user that isn't allowed to edit the ticket status:
image

Same for the private toggle: 

image

And the followup source + template dropdowns.

"Answer" dropdown does no do anything:
image

btry:
Hi

# CSS problem => solved
image

all other problems occur when I access a ticket with Formcreator disabled, then I think they need a fix in GLPI core.

AdrienClairembault:
Ok, I'll work on fixing these issues in GLPI core before testing again this PR then.

AdrienClairembault:
PR for readonly status alignement: https://github.com/glpi-project/glpi/pull/10253

AdrienClairembault:
PR hiding extra actions : https://github.com/glpi-project/glpi/pull/10256

AdrienClairembault:
Not sure if this is a core issue or not but the number of linked items show -1 instead of 1:

image

Left border is missing:

image

Also, not related to this PR as this also happen on the develop branch but it's impossible to search for forms because of a sql error:

```
[2022-01-04 10:07:08] glpisqllog.ERROR: DBmysql::query() in /home/adrie/localhost/gmaster/src/DBmysql.php line 352
  *** MySQL query error:
  SQL: SELECT DISTINCT `glpi_plugin_formcreator_forms`.`id` AS id, 'glpi' AS currentuser,
                        `glpi_plugin_formcreator_forms`.`entities_id`, `glpi_plugin_formcreator_forms`.`is_recursive`,  `glpi_plugin_formcreator_forms`.`name` AS `ITEM_PluginFormcreatorForm_1`,
                        `glpi_plugin_formcreator_forms`.`id` AS `ITEM_PluginFormcreatorForm_1_id`,
                        `glpi_plugin_formcreator_forms`.`is_active` AS `ITEM_PluginFormcreatorForm_30`,  `glpi_plugin_formcreator_categories`.`name` AS `ITEM_PluginFormcreatorForm_10`,  `glpi_plugin_formcreator_forms`.`language` AS `ITEM_PluginFormcreatorForm_7`,  `glpi_plugin_formcreator_forms`.`helpdesk_home` AS `ITEM_PluginFormcreatorForm_8`,  `glpi_plugin_formcreator_forms`.`access_rights` AS `ITEM_PluginFormcreatorForm_9` FROM `glpi_plugin_formcreator_forms`LEFT JOIN `glpi_plugin_formcreator_categories` 
                                          ON (`glpi_plugin_formcreator_forms`.`plugin_formcreator_categories_id` = `glpi_plugin_formcreator_categories`.`id`
                                              ) WHERE  `glpi_plugin_formcreator_forms`.`is_deleted` = 0  ORDER BY `ITEM_PluginFormcreatorForm_1` ASC  LIMIT 0, 20
  Error: Table 'glpi-master.glpi_plugin_formcreator_categories' doesn't exist
  Backtrace :
  src/Search.php:1369                                
  src/Search.php:362                                 Search::constructData()
  src/Search.php:114                                 Search::getDatas()
  src/Search.php:97                                  Search::showList()
  plugins/formcreator/front/form.php:50              Search::show()
```

AdrienClairembault (changes requested):
Left border is still missing (see previous comment).

The "cancel my ticket" button is displayed incorrectly (no "pointer" cursor, different height that the previous button and maybe lacking an icon ?)

image

Also what is this button supposed to do ?
From what I understand it put the issue in the trash bin (is_deleted = 1).
Isn't it the same as the thrash can icon on the bottom right corner ?
Maybe we should only keep the current trashcan button but replace its tooltip by "Cancel my ticket" ?

Also, once the ticket is in the trash can it still show up in my tickets list:
image

image

AdrienClairembault (approved):

PR #2467, opened 2021-11-19
feat: compliance with GLPI PR#9898

Rom1-B:
### Changes description

Compliance with PR9898 in GLPI 10
[](https://github.com/glpi-project/glpi/pull/9898)

- Allow multiple categories to be added to a knowledge base article.
- Allow access to a knowledge base article by self-service users without it being in the FAQ.

### Checklist

Please check if your PR fulfills the following specifications:

- [ ] Tests for the changes have been added
- [ ] Docs have been added/updated

### References

Closes #N/A

PR #2502, opened 2021-12-15
Category tree broken

btry:
#2467 is incomplete, browsing categories is broken

### DIFF
--- a/ajax/homepage_wizard.php
+++ b/ajax/homepage_wizard.php
@@ -53,7 +53,7 @@ if ($_REQUEST['wizard'] == 'categories') {
 }
 
 function plugin_formcreator_showWizardCategories() {
-   $tree = PluginFormcreatorCommon::getCategoryTree(0, false);
+   $tree = PluginFormcreatorCommon::getCategoryTree();
    echo json_encode($tree, JSON_UNESCAPED_SLASHES);
 }
 
--- a/ajax/homepage_wizard.php
+++ b/ajax/homepage_wizard.php
@@ -53,7 +53,7 @@ if ($_REQUEST['wizard'] == 'categories') {
 }
 
 function plugin_formcreator_showWizardCategories() {
-   $tree = PluginFormcreatorCategory::getCategoryTree(0, false);
+   $tree = PluginFormcreatorCommon::getCategoryTree(0, false);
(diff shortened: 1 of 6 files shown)


Rom1-B:
Text is cut off in the list:

image

Rom1-B (commented):
When setting the "merged with forms" entity, there is an error with "see all":
image
But it works with the "distinct menu entry" parameter.

ajax/homepage_wizard.php
```diff
@@ -53,7 +53,7 @@
 }
 
 function plugin_formcreator_showWizardCategories() {
-   $tree = PluginFormcreatorCategory::getCategoryTree(0, false);
+   $tree = PluginFormcreatorCommon::getCategoryTree(0, false);
```
Rom1-B:
```suggestion
   $tree = PluginFormcreatorCommon::getCategoryTree(0, false);
```
Maybe change also in `ajax/kb_category.php`

AdrienClairembault:
If I understand correctly, form categories have been removed and we now use only kb categories.
It seems unusual to have "Forms" in a "Knowledge base category".

Shouldn't the item be renamed to something more generic ("Category"  ?).

AdrienClairembault (approved):
Functionally seems OK, category tree is working correctly.

btry:
The service catalog was designed to merge FAQ and forms, helping the user to find a solution to his problem from FAQ before filling an assistance request. Then dropping form categories and the not very convenient relation between form categories and KB categories seems to me a good thing. 

I think "category" would be to much generic (one could ask "category of what ?"), and this change would be in GLPI core, might break some other plugin.

AdrienClairembault:
> The service catalog was designed to merge FAQ and forms, helping the user to find a solution to his problem from FAQ before filling an assistance request. Then dropping form categories and the not very convenient relation between form categories and KB categories seems to me a good thing.

Yes this is a good change 👍 .

> I think "category" would be to much generic (one could ask "category of what ?"), and this change would be in GLPI core, might break some other plugin.

I was thinking of changing the displayed name only, not the class name.
Shouldn't break anything.
I agree "Category" is a very generic... how about something like "Helpdesk category" ?

It just seems badly named right now as it does not contain only Knowledge base items.

PR #2509, opened 2021-12-21
Check internal dependencies only before activation

btry:
in some cases, it may be impossible to clean, install or upgrade the plugin becausee of prerequisite checks.


AdrienClairembault:
Isn't it risky to allow installing the plugin without dependencies ?
Some upgrade code might use vendors code.

Not mandatory but if you want you can also use `$plugin->isInstalled('formcreator') && !$plugin->isActivated('formcreator')` as an alternative condition.

PR #2516, opened 2021-12-24
Enhance skip checks flag implementation

btry:
the flag skip_checks may be set by end user, it is better to avoid this

### DIFF
--- a/inc/condition.class.php
+++ b/inc/condition.class.php
@@ -372,9 +372,11 @@ class PluginFormcreatorCondition extends CommonDBChild implements PluginFormcrea
       foreach ($sections as $section) {
          $sectionsList[] = $section->getID();
       }
-      $questionListExclusion[] = [
-         PluginFormcreatorSection::getForeignKeyField() => $sectionsList,
-      ];
+      if (count($sectionsList)) {
+         $questionListExclusion[] = [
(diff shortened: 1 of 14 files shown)


inc/condition.class.php
```diff
@@ -372,9 +372,11 @@ public function getConditionHtml(array $input) : string {
...
          $sectionsList[] = $section->getID();
       }
-      $questionListExclusion[] = [
-         PluginFormcreatorSection::getForeignKeyField() => $sectionsList,
-      ];
+      if (count($sectionsList)) {
+         $questionListExclusion[] = [
+            PluginFormcreatorSection::getForeignKeyField() => $sectionsList,
+         ];
+      }
```
btry, line 379: extra fix : avoid an empty IN expression in SQL query.

inc/conditionnabletrait.class.php
```diff
@@ -76,18 +76,16 @@ public function checkConditions(array $input): bool {
       return true;
    }
 
-   public function updateConditions($input) : bool {
+   public function checkConditionSettings(array $input): bool {
```
btry, line 79: data integrity checks for conditions are now passive : no more change in the DB. IT is the responsibility of the itemtype to set show_rule accordying to the data integrity check in prepareInputForUpdate().

AdrienClairembault (approved):

PR #2519, opened 2021-12-27
PluginFormcreatorForm::showForm() Use twig 

btry:
use Twig in  PluginFormcreatorForm::showForm() 

Maybe merge this PR with all others twig conversions.


front/form.form.php
```diff
@@ -41,8 +41,12 @@
...
    // Add a new Form
    Session::checkRight('entity', UPDATE);
-   $newID = $form->add($_POST);
-   Html::redirect(FORMCREATOR_ROOTDOC . '/front/form.form.php?id=' . $newID);
+   if ($newID = $form->add($_POST)) {
+     if ($_SESSION['glpibackcreated']) {
+         Html::redirect($computer->getLinkURL());
+      }
+   }
+   Html::back();
```
btry, line 49: extra : better rediretion

ajax/question_add.php
```diff
@@ -44,7 +44,6 @@
     http_response_code(400);
     exit;
 }
-$question->getFromDB($question->getID()); // To remove if GLPI #9210 merged
```
btry, line 47: the mentionned PR is merged

ajax/question_delete.php
```diff
@@ -39,6 +39,7 @@
 $questionId = (int) $_REQUEST['id'];
 $question = new PluginFormcreatorQuestion();
 if (!$question->getFromDB($questionId)) {
+    echo __('Question not found', 'formcreator');
```
btry, line 42: if deletion fails, no error displayed, fixing it

css/styles.scss
```diff
@@ -93,9 +93,15 @@ html, body {
  * Form designer
  */
 
+/* modals */
+.modal-dialog .card-header {
+   // Fix ribbon position
+   position: relative;
+}
+
```
btry, line 101: fix ribbon position when showForm() is used in a modal

css/styles.scss
```diff
@@ -93,9 +93,15 @@ html, body {
...
 /* Sections */
 #plugin_formcreator_form.plugin_formcreator_form_design {
-   > ol {
```
btry, line 98: less strict CSS to match the expected elements under GLPI 10

front/form.form.php
```diff
@@ -41,8 +41,12 @@
...
    // Add a new Form
    Session::checkRight('entity', UPDATE);
-   $newID = $form->add($_POST);
-   Html::redirect(FORMCREATOR_ROOTDOC . '/front/form.form.php?id=' . $newID);
+   if ($newID = $form->add($_POST)) {
+      if ($_SESSION['glpibackcreated']) {
+         Html::redirect($form->getLinkURL());
+      }
+   }
+   Html::back();
```
btry, line 49: obey $_SESSION['glpibackcreated'] preference

inc/common.class.php
```diff
@@ -158,7 +158,9 @@ public static function prepareBooleanKeywords(string $keywords) : string {
     * @return array
     */
    public static function getFontAwesomePictoNames() : array {
-      $list = require_once(GLPI_PLUGIN_DOC_DIR . '/formcreator/' . self::getPictoFilename(GLPI_VERSION));
+      static $list = null;
+
+      $list = $list ?? require_once(GLPI_PLUGIN_DOC_DIR . '/formcreator/' . self::getPictoFilename());
```
btry, line 163: avoid failure if the require statement is ran twice

inc/common.class.php
```diff
@@ -168,16 +170,16 @@ public static function getFontAwesomePictoNames() : array {
     * @param $version string GLPI version
     * @return string
     */
-   public static function getPictoFilename(string $version) : string {
```
btry, line 171: obsolete argument removed

inc/condition.class.php
```diff
@@ -224,7 +226,7 @@ public function export(bool $remove_uuid = false) : array {
     * @param CommonDBTM $item
     * @return PluginFormcreatorCondition[]
     */
-   public function getConditionsFromItem(CommonDBTM $item) : array {
+   public static function getConditionsFromItem(CommonDBTM $item) : array {
```
btry, line 229: make method callable from a template

inc/condition.class.php
```diff
@@ -249,6 +251,14 @@ public function getConditionsFromItem(CommonDBTM $item) : array {
       return $conditions;
    }
 
+   public static function countForItem(PluginFormcreatorConditionnableInterface $item): int {
```
btry, line 254: make countElementsInTable callable from a template

inc/condition.class.php
```diff
@@ -276,153 +286,96 @@ public function showConditionsForItem(PluginFormcreatorConditionnableInterface $
...
-               'value'                 => $show_logic,
-               'display_emptychoice'   => false,
-               'rand'                  => $rand,
-            ]);
-      $html.= '</div>';
-
-      // dropdown of questions
-      $form = PluginFormcreatorCommon::getForm();
-      $questionListExclusion = [];
+   public static function getQuestionsExclusion(PluginFormcreatorConditionnableInterface $item) {
```
btry, line 301: isolate in dedicated method the code to exclude questions from a questions dropdown

inc/field/actorfield.class.php
```diff
@@ -249,11 +249,7 @@ protected function sanitizeValue($value): array {
             $user->getFromDB($item);
             if (!$user->isNewItem()) {
                // A user known in the DB
-               if (method_exists($user, 'getFriendlyName')) {
-                  $knownUsers[$user->getID()] = $user->getFriendlyName();
-               } else {
-                  $knownUsers[$user->getID()] = $user->getRawName();
-               }
+               $knownUsers[$user->getID()] = $user->getFriendlyName();
```
btry, line 252: remove backward compatibility with GLPI 9.5

inc/fieldinterface.class.php
```diff
@@ -291,4 +291,6 @@ public function isVisibleField() : bool;
     * @return array translatable strings under keys 'string' and 'text'
     */
    public function getTranslatableStrings(array $options = []) : array;
+
+   public function getQuestion();
```
btry, line 295: fix an IDE warning

AdrienClairembault (changes requested):

AdrienClairembault:
If I remember correctly; this option was set to 'Yes' by default in the previous version:

image

-> This option was requested as an edge case, most people will want it set to 'Yes' for all their forms.

AdrienClairembault:
Visual issues when creating a section:

image

AdrienClairembault:
After adding my section I can't add any question, the tab has an error: 
image

ajax/question_delete.php
btry: The error mesage is processed by js/scripts.js. Your sgestion is possible, but also requires an update in the JS  code. This PR is complicated enough, then we may do this in a separate enhancement PR.

AdrienClairembault:
Adding an actor to a target doesn't seem to work:

image

After save:

image

AdrienClairembault:
I'm unable to edit the target conditions:

image

AdrienClairembault:
This buttons isn't doing anything:

image

I suppose it should toggle some search criteria ?

AdrienClairembault:
Visual issues when adding a new language to my form:

image

btry:
> Visual issues when trying to add a category from the "New item - Form" page:

this page is not yet converted into Twig. 

> Adding a validator to my form doens't seem to work

Fixed 

> Visual issues when creating a section (entity name + child entity checkbox):

Even when isEntityAssign returns false, the UI shows them. I tried to design templates as generic as possible, by using core's templates. The elements related to the entity are disabled, but I agree it would be better ti not show them. To do thos, it may be better to fix the template in the core.

About the empty ribbon : I did not focused on that for now. I find it not disturbing as is. Defining the icon for this itemtype should solve the topic. This is also true for question itemtype.

btry:
> Adding an actor to a target doesn't seem to work:

fixed

btry:
image

The toggle button is always displayed by GLPI. i think this page should not show the search filters, then GLPI should provide some mechanism to disable this toggle button.

btry:
ribbon in form language tab : 

needs 

```scss
.card_header {
   position:relative
}
```

This would also fix the other identified cases, but needs to be evaluated before inclusion in GLPI core.

AdrienClairembault:
> Even when isEntityAssign returns false, the UI shows them. I tried to design templates as generic as possible, by using core's templates. The elements related to the entity are disabled, but I agree it would be better ti not show them. To do thos, it may be better to fix the template in the core.

Done -> https://github.com/glpi-project/glpi/pull/10275.

AdrienClairembault (changes requested):
## UI Stuffs / Subjective

1\) Maybe we can add a few improvement to this page ?
image
image

The input move when you switch between group and users, maybe a vertical layout would use the available space better (not very important).

2\) Maybe add some margin before displaying the first section:
image
image

Also the dark grey backgrounds seems a bit out of place in the new UI.
Maybe there is some tabler component we would use here but I'm not sure which.


## Bugs

1\) If I click on this dropdown, the modal close itself and my unsaved question is lost:
image

Seems to behave differently depending on the question type.
Textarea closes the dropdown immediatly but some others don't.

2\) If I switch between all the possible values, new inputs keep popping up below the form:
image

3\) If the section isn't saved yet there is never any input however:
image

4\) Same as 2 but for the submit button conditions:
image

5\) Same as 3 but for a new question, can't add any condition:
image

6\) When creating a question, submit button doesn't close the modal.
Each time I press the button another question is added in the background.
Also can't close the modal.
image

7\) Same as 2 but when editing a question:
image

8\) Not sure if caused by the twig changes but I can submit some partial values for targets (no question selected):
image

9\) Same as 2 but for target condition:
image

10\) Can't add validator as a requester for a target change or problem:
image
After submitting the value isn't added.
It does work for target ticket, I'm even able to add the user multiple time (is this ok ?):
image

11\) Actor questions aren't visible in the preview:
image
image

12\) Saving a question doesn't close the modal, can't close with the exit button either.

AdrienClairembault:
Another weird tab bug that I've been able to reproduce:
If you go to a form and:
- Go to "Targets" tab
- Go to "Access types" tabs
- Go back to "Targets" tab
- Add a new target
-> On submit, you are redirected to the "Access types" tab instead of staying on the "Targets" tab

btry:
I just noticed similar behaviours elsewhere. I'll fix all those things next week.

RoboFile.php
```diff
@@ -300,18 +300,7 @@ public function updateChangelog() {
...
-      $command = "xgettext $phpSources -o locales/$potfile -L PHP --add-comments=TRANS --from-code=UTF-8 --force-po";
-      $command.= " --keyword=_n:1,2,4t --keyword=__s:1,2t --keyword=__:1,2t --keyword=_e:1,2t --keyword=_x:1c,2,3t --keyword=_ex:1c,2,3t";
-      $command.= " --keyword=_sx:1c,2,3t --keyword=_nx:1c,2,3,5t";
-      $this->_exec($command);
-      $jsSources = "js/*.js";
-      $command = "xgettext $jsSources -o locales/$potfile -j -L javascript --add-comments=TRANS --from-code=UTF-8 --force-po";
-      $command.= " --keyword=_n:1,2,4t --keyword=__s:1,2t --keyword=__:1,2t --keyword=_e:1,2t --keyword=_x:1c,2,3t --keyword=_ex:1c,2,3t";
-      $command.= " --keyword=_sx:1c,2,3t --keyword=_nx:1c,2,3,5t";
-      $this->_exec($command);
+      $this->_exec('tools/extract_template.sh');
```
btry, line 303: As Robo is being abandonned in GLPI project, I pick the locales extraction tool from GLPI, to find strings from twig templates

ajax/question_add.php
```diff
@@ -41,15 +41,15 @@
 }
 
 if (!$question->add($_REQUEST)) {
-    http_response_code(400);
+    http_response_code(500);
```
btry, line 44: err 500 is more appropriate here

ajax/condition.php
```diff
@@ -32,6 +32,31 @@
...
+      die();
+   }
+}
+
 // get an empty condition HTML table row
 $condition = new PluginFormcreatorCondition();
-echo $condition->getConditionHtml($_POST);
+$condition->fields['itemtype'] = $_POST['itemtype'];
+$condition->fields['items_id'] = (int) $_POST['items_id'];
+echo $condition->getConditionHtml($parent);
```
btry, line 62: Atempt to simplify conditions creation and display for all conditionnable items. As conditions may be added to non-existing items (creating items and conditions at the same time), we need to forge items to make things work properly.

In other cases, loading them from DB is fine.

The most important need is to be able to find the form on top of the items, which are themselve parent of the conditions to show.

ajax/question_add.php
```diff
@@ -41,15 +41,15 @@
...
-$question->getFromDB($question->getID()); // To remove if GLPI #9210 merged
 $json = [
     'y'      => $question->fields['row'],
     'x'      => $question->fields['col'],
     'width'  => $question->fields['width'],
     'height' => '1',
     'html'   => $question->getDesignHtml(),
 ];
-echo json_encode($json);
+echo json_encode($json, JSON_UNESCAPED_UNICODE);
```
btry, line 55: usually, we shoud should prevent unicode chars escaping

install/mysql/plugin_formcreator_empty.sql
```diff
@@ -195,42 +195,42 @@ CREATE TABLE IF NOT EXISTS `glpi_plugin_formcreator_targetchanges` (
...
+  `commonitil_validation_rule`     int(11) NOT NULL DEFAULT '1',
+  `commonitil_validation_question` varchar(255) DEFAULT NULL,
+  `show_rule`                      int(11) NOT NULL DEFAULT '1',
+  `sla_rule`                       int(11) NOT NULL DEFAULT '1',
+  `sla_question_tto`               int(11) NOT NULL DEFAULT '0',
+  `sla_question_ttr`               int(11) NOT NULL DEFAULT '0',
+  `ola_rule`                       int(11) NOT NULL DEFAULT '1',
+  `ola_question_tto`               int(11) NOT NULL DEFAULT '0',
+  `ola_question_ttr`               int(11) NOT NULL DEFAULT '0',
+  `uuid`                           varchar(255) DEFAULT NULL,
```
btry, line 233: better indentation

js/scripts.js
```diff
@@ -464,8 +464,6 @@ var plugin_formcreator = new function() {
             $(item.ddElement.el).find('a').off('click.prevent');
          },
          300);
-         // Remove empty rows
-         plugin_formcreator.moveUpItems(group);
```
btry, line 468: the method is disabled, calling it is useless

js/scripts.js
```diff
@@ -548,7 +546,8 @@ var plugin_formcreator = new function() {
   this.startChangeItem = function (event, item) {
       $(item.ddElement.el).find('a').on('click.prevent', function(event) {
          return false;
-      });      var items = $(event.currentTarget).find('> .grid-stack-item');
+      });
+      var items = $(event.currentTarget).find('> .grid-stack-item');
```
btry, line 550: style problem here

js/scripts.js
```diff
@@ -597,6 +596,7 @@ var plugin_formcreator = new function() {
          plugin_formcreator.dirty = false;
       }).done(function(response) {
          plugin_formcreator.dirty = false;
+         that.resetTabs();
```
btry, line 599: when qhestions are moved, reset other tabs of the form to force reload of preview when the admin wants to test the form

js/scripts.js
```diff
@@ -624,8 +624,8 @@ var plugin_formcreator = new function() {
    // === QUESTIONS ===
 
    this.deleteQuestion = function (target) {
-      var item = $(target).closest('.grid-stack-item');
-      var id = item.attr('data-id');
+      var item = target.closest('.grid-stack-item');
+      var id = item.getAttribute('data-id');
```
btry, line 628: switch to vanilla JS when possible

js/scripts.js
```diff
@@ -611,7 +611,7 @@ var plugin_formcreator = new function() {
          if (id != that.changingItemId) {
             return;
          }
-         $(event.target).gridstack.update(
+         event.target.gridstack.update(
```
btry, line 614: vanilla JS when possible

js/scripts.js
```diff
@@ -643,29 +643,10 @@ var plugin_formcreator = new function() {
...
       }
    };
 
-   /**
-    * Move up items in a grid when row is empty
-    * @param grid stack container
-    * @param row  row to fill with items after it
-    */
-   this.moveUpItems = function (grid) {
-      return;
```
btry, line 657: method was disabled (never  used in any release), dropping it

js/scripts.js
```diff
@@ -697,17 +678,18 @@ var plugin_formcreator = new function() {
    }
 
    this.addQuestion = function () {
-      var form = $('form[data-itemtype="PluginFormcreatorQuestion"]');
+      var form = document.querySelector('form[data-itemtype="PluginFormcreatorQuestion"]');
```
btry, line 681: vanilla JS if possible

js/scripts.js
```diff
@@ -697,17 +678,18 @@ var plugin_formcreator = new function() {
...
       var that = this;
-      $.ajax({
+      $.post({
          url: formcreatorRootDoc + '/ajax/question_add.php',
-         type: "POST",
-         data: form.serializeArray(),
-         dataType: 'json'
+         data: new FormData(form),
+         processData: false,
+         contentType: false,
```
btry, line 687: processData and contentType are here to help JQuery to work with FormData object; hels to remove useless JQuery use

js/scripts.js
```diff
@@ -697,17 +678,18 @@ var plugin_formcreator = new function() {
...
+         dataType: 'json',
       }).fail(function(data) {
          displayAjaxMessageAfterRedirect();
       }).done(function(data) {
-         var sectionId = form.find('select[name="plugin_formcreator_sections_id"]').val();
+         var sectionId = form.querySelector('select[name="plugin_formcreator_sections_id"]').value;
```
btry, line 692: less JQuery

js/scripts.js
```diff
@@ -726,30 +708,29 @@ var plugin_formcreator = new function() {
...
       var form = $('form[data-itemtype="PluginFormcreatorQuestion"]');
       var questionId = form.find('[name="id"]').val();
       var that = this;
-      $.ajax({
+      $.post({
          url: formcreatorRootDoc + '/ajax/question_update.php',
-         type: "POST",
          data: form.serializeArray(),
-         dataType: 'html'
+         dataType: 'json'
```
btry, line 720: returned body switched to JSON

js/scripts.js
```diff
@@ -726,30 +708,29 @@ var plugin_formcreator = new function() {
...
       }).fail(function(data) {
          displayAjaxMessageAfterRedirect();
       }).done(function(data) {
          var question = $('.plugin_formcreator_form_design[data-itemtype="PluginFormcreatorForm"] [data-itemtype="PluginFormcreatorQuestion"][data-id="' + questionId + '"]');
-         question.find('[data-field="name"]').text(data)
+         question.find('[data-field="name"]').text(data['name'])
```
btry, line 725: returned body switched to JSON

js/scripts.js
```diff
@@ -726,30 +708,29 @@ var plugin_formcreator = new function() {
...
          that.resetTabs();
       });
+
    }
 
    this.duplicateQuestion = function (target) {
-      var item = $(target).closest('.grid-stack-item');
-      var id = item.attr('data-id');
+      var item = target.closest('.grid-stack-item');
+      var id = item.getAttribute('data-id');
```
btry, line 733: less JQuery

js/scripts.js
```diff
@@ -888,13 +869,28 @@ var plugin_formcreator = new function() {
...
+                  questionId = idInput.getAttribute('value');
+               }
+               if (questionId === null) {
+                  plugin_formcreator.addQuestion(event);
+               } else {
+                  plugin_formcreator.editQuestion(event);
+               }
+               $('#' + modalId).modal('hide');
+            });
+         }
```
btry, line 893: fix non closing modals.

js/scripts.js
```diff
@@ -922,52 +918,76 @@ var plugin_formcreator = new function() {
...
+                  sectionId = idInput.getAttribute('value');
+               }
+               if (sectionId === null) {
+                  plugin_formcreator.addSection(event);
+               } else {
+                  plugin_formcreator.editSection(event);
+               }
+               $('#' + modalId).modal('hide');
+            });
+         }
```
btry, line 943: fix non closing modal

js/scripts.js
```diff
@@ -922,52 +918,76 @@ var plugin_formcreator = new function() {
...
-         type: "POST",
-         data: form.serializeArray(),
-         dataType: 'html'
-      }).fail(function(data) {
-         alert(data.responseText);
-      }).done(function(data) {
+         processData: false,
+         contentType: false,
+         data: new FormData(form),
+         dataType: 'json'
```
btry, line 955: switch to json response

js/scripts.js
```diff
@@ -922,52 +918,76 @@ var plugin_formcreator = new function() {
...
+               }
+               $('#' + modalId).modal('hide');
+            });
+         }
       });
    }
 
-   this.addSection = function () {
-      var form = $('form[data-itemtype="PluginFormcreatorSection"]');
+   this.addSection = function (event) {
```
btry, line 947: event pssed to function, helping to mot search too much in the DOM

js/scripts.js
```diff
@@ -922,52 +918,76 @@ var plugin_formcreator = new function() {
...
+         $(myModal).modal('hide');
       });
-      glpi_close_all_dialogs();
+
    }
 
-   this.editSection = function () {
-      var form = $('form[data-itemtype="PluginFormcreatorSection"]');
-      var sectionId = form.find('[name="id"]').val();
+   this.editSection = function (event) {
```
btry, line 972: event passed to function, here as well

js/scripts.js
```diff
@@ -1122,6 +1148,31 @@ var plugin_formcreator = new function() {
       });
    }
 
+   this.changeActor = function(type, value) {
```
btry, line 1151: factorization of :
plugin_formcreator_ChangeActorRequester
plugin_formcreator_ChangeActorWatcher
plugin_formcreator_ChangeActorASsigned

js/scripts.js
```diff
@@ -1313,81 +1388,6 @@ function plugin_formcreator_hideAssignedForm() {
    $('#btn_cancel_assigned').hide();
 }
 
-function plugin_formcreator_ChangeActorRequester(value) {
```
btry, line 1316: factorized

js/scripts.js
```diff
@@ -1313,81 +1388,6 @@ function plugin_formcreator_hideAssignedForm() {
...
-      case '6' : $('#block_requester_question_group').show();         break;
-      case '10': $('#block_requester_group_from_object').show();      break;
-      case '11': $('#block_requester_tech_group_from_object').show(); break;
-      case '9' : $('#block_requester_question_actors').show();        break;
-      case '7' : $('#block_requester_supplier').show();               break;
-      case '8' : $('#block_requester_question_supplier').show();      break;
-   }
-}
-
-function plugin_formcreator_ChangeActorWatcher(value) {
```
btry, line 1340: factorized

js/scripts.js
```diff
@@ -1313,81 +1388,6 @@ function plugin_formcreator_hideAssignedForm() {
...
-      case '9' : $('#block_watcher_question_actors').show();        break;
-      case '10': $('#block_watcher_group_from_object').show();      break;
-      case '11': $('#block_watcher_tech_group_from_object').show(); break;
-      case '9' : $('#block_watcher_question_actors').show();        break;
-      case '7' : $('#block_watcher_supplier').show();               break;
-      case '8' : $('#block_watcher_question_supplier').show();      break;
-   }
-}
-
-function plugin_formcreator_ChangeActorAssigned(value) {
```
btry, line 1365: factorized

js/scripts.js
```diff
@@ -1169,6 +1220,41 @@ var plugin_formcreator = new function() {
          }
       );
    }
+
+   this.deleteActor = function (item) {
```
btry, line 1224: convert add / delete of target actors calls into Ajax, for better UX.

js/scripts.js
```diff
@@ -1183,17 +1269,6 @@ function plugin_formcreator_addTarget(items_id) {
...
-function plugin_formcreator_editTarget(itemtype, items_id) {
-   glpi_ajax_dialog({
-      dialogclass: 'modal-xl',
-      url: formcreatorRootDoc + '/ajax/target_edit.php',
-      params: {
-         itemtype: itemtype,
-         id: items_id
-      },
-   });
-}
```
btry, line 1195: no longer used

js/scripts.js
```diff
@@ -1517,32 +1517,26 @@ function plugin_formcreator_toggleCondition(target) {
...
+      contentType: false,
+      data: data,
+   }).done(function (data) {
+      target.closest('div.row').after(document.createRange().createContextualFragment(data));
    });
 }
 
 function plugin_formcreator_removeNextCondition(target) {
-   $(target).parents('tr').remove();
+   target.closest('div.row').remove();
```
btry, line 1539: update to new DOM, drop jquery usage

tests/3-unit/PluginFormcreatorCondition.php
```diff
@@ -35,7 +35,7 @@ class PluginFormcreatorCondition extends CommonTestCase {
    public function beforeTestMethod($method) {
       parent::beforeTestMethod($method);
 
-      self::login('glpi', 'glpi');
+      $this->login('glpi', 'glpi');
```
btry, line 38: the methid is not static

tests/3-unit/PluginFormcreatorCondition.php
```diff
@@ -82,14 +82,14 @@ public function testGetConditionsFromItem() {
       $this->boolean($condition->isNewItem())->isFalse();
 
       // Check that all conditions are retrieved
-      $output = $condition->getConditionsFromItem($question);
```
btry, line 85: method is now static

btry, line 85: method is now static

tests/3-unit/PluginFormcreatorCondition.php
btry, line 38: the methid is not static

tests/3-unit/PluginFormcreatorCondition.php
```diff
@@ -82,14 +82,14 @@ public function testGetConditionsFromItem() {
...
+      $output = \PluginFormcreatorCondition::getConditionsFromItem($question);
       $this->array($output)->hasSize(2);
    }
 
    public function testImport() {
       $question = $this->getQuestion();
       $form = new \PluginFormcreatorForm();
-      $form->getFromDBByQuestion($question);
+      $form = \PluginFormcreatorForm::getByItem($question);
```
btry, line 92: use a more generic method to find a form from an item inside it

tests/3-unit/PluginFormcreatorCondition.php
```diff
@@ -144,11 +144,9 @@ public function testExport() {
       ]);
       $question2 = $this->getQuestion([
          'plugin_formcreator_forms_id' => $form->getID(),
-      ]);
-      $question2->updateConditions([
```
btry, line 148: adding conditions can be done while adding the question

inc/form.class.php
```diff
@@ -2198,52 +1956,57 @@ public function isPublicAccess() : bool {
...
-    * gets the form containing the given section
+    * Get a form a sub object (section, question, target)
+    * Able to find a form from an item having only its parent FK and no ID
     *
-    * @param PluginFormcreatorSection $section
-    * @return boolean true if success else false
+    * @param CommonDBTM $item
+    * @return null|self
     */
-   public function getFromDBBySection(PluginFormcreatorSection $item) {
```
btry, line 2206: replaced by more generic getByItem method

inc/form.class.php
```diff
@@ -2198,52 +1956,57 @@ public function isPublicAccess() : bool {
...
-   public function getFromDBByTarget(CommonDBTM $item) {
-      if ($item->isNewItem()) {
-         return false;
+      if ($item::getType() == self::getType()) {
+         return $item;
       }
-      return $this->getFromDB($item->fields[self::getForeignKeyField()]);
-   }
 
-   public function getFromDBByQuestion(PluginFormcreatorQuestion $question) {
```
btry, line 2220: replaced by more generic getByItem method

inc/question.class.php
```diff
@@ -755,6 +739,14 @@ public function post_purgeItem() {
    }
 
    public function showForm($ID, $options = []) {
+      // $options['candel'] = false;
+      // $options['formoptions'] = sprintf('data-itemtype="%s"', self::getType());
+      // TemplateRenderer::getInstance()->display('@formcreator/pages/question.html.twig', [
+      //    'item' => $this,
+      //    'params' => $options,
+      // ]);
+      // return true;
```
btry, line 748: too much complex to convert now, lets keep this draft and continue in an other PR

inc/section.class.php
```diff
@@ -86,6 +101,12 @@ public function rawSearchOptions() {
       return $tab;
    }
 
+   public function getForbiddenStandardMassiveAction() {
+      return [
+         'update', 'clone', 'add_note',
+      ];
+   }
```
btry, line 108: forbid some massive actions

inc/target_actor.class.php
```diff
@@ -92,6 +92,39 @@ public static function getTypeName($nb = 0) {
    }
 
    public function prepareInputForAdd($input) {
+      $requiredKeys = ['itemtype', 'items_id', 'actor_role', 'actor_type'];
```
btry, line 95: add integrity checks, does not handles all cases for now, but fix the inconsistency found this week

inc/targetproblem.class.php
```diff
@@ -669,6 +558,58 @@ public static function countItemsToImport(array $input) : int {
       return 1 + self::countChildren($subItems, $input);
    }
 
+   public function defineTabs($options = []) {
```
btry, line 561: like target ticket and target change : swtch to a multiple tabs design to edit this target

inc/targetticket.class.php
```diff
@@ -794,20 +697,6 @@ public function pre_purgeItem() {
...
-         $this->updateConditions($this->input);
-      }
-   }
-
-   public function post_updateItem($history = 1) {
-      parent::post_updateItem();
-      if ($this->input['show_rule'] != PluginFormcreatorCondition::SHOW_RULE_ALWAYS) {
-         $this->updateConditions($this->input);
-      }
-   }
```
btry, line 809: factorize with all targets types, see implementation in AbstractTarget. 

If adding actors by default is not wanted, then create for each class a method to add default actors.

inc/targetticket.class.php
```diff
@@ -1144,7 +1033,7 @@ protected  function showTypeSettings($rand) {
       echo '</div>';
       echo '<div id="requesttype_question_value" style="display: none">';
       PluginFormcreatorQuestion::dropdownForForm(
```
btry, line 1035: dropdownForForm now gets form object as 1st arg

AdrienClairembault (changes requested):
1\) The visibility condition dropdown is too large, it doesn't seem to fit the input:
image

Also, if you have a scroll-able page (a big question form like the checkbox question type for example), the options are disconnected from the input:

image

2\) When creating a form, "All langages" is displayed differently when selected vs the value in the dropdown:
image
  
Also it should be  "All lang**u**ages" instead of  "All langages".

3\) I'm not sure what happened but the first form I tried to create didn't had the correct default values.
The visible option was on "No", the default language on "All Languages" and both colors to white #FFFFFF.
After submitting the form it was also not saved in the database instead it was used to modify an existing form:
image
I'm 100% sure of it as it was the exact colors and icons I had picked + the correct time.

Every other form I tried to create after this didn't have the issue so I wasn't able to reproduce this bug.

UPDATE, I've been able to reproduce it: 
- Save the questions on an existing form
- Try to create a new form

image

4\) The validation input still "moves around" depending on the selected value
image
It should use the new bootstrap col/row forms instead of the legacy html table.

The label "Need validaton?" has a spelling mistake.


5\) If I click twice on the "move section down" button (which I first though was a button to hide the section content, not move my section. not sure how to improve this UI wise) then some questions are not shown anymore.

Before:
image

After: 
image


6\) The submit button visibility condition trigger an error:
image

You need to set a condition and save, the error wont be shown if no conditions are set.

Seems to be OK for questions and sections.

7\) I can't update my target ticket, right errors.
image

This is a GLPI with only formcreator and no rights / configuration modification.

 8\) Two issues with the actor fields:
image

There is an excessive margin compared to similar inputs used by GLPI:
image

Also the default "----" option shouldn't be selectable (might be fixed by https://github.com/glpi-project/glpi/pull/10288)

9\) I can't click on these icons in a form target, seems to do nothing:
image

Beside not having a "pointer" cursor, the 'delete' button seems to trigger a 404 error:
image
The "notification" button doesn't seem to do anything.

Same 404 error if I try to add another actor.

10\) The "issues" button lead me to an error:

image
image

The issue seems to be that I'm able to add a "Tag" question to my forms without having the tag plugin installed or activated.
This also prevent me from displaying this form.

I think firstly I shouldn't be able to add a Tag question if I don't have the plugin activated.
Even tho, this kind of case might still happen (If I enable tag, create a tag question then disable the tag plugin for example) so we should avoid throwing an error and just skip the question.

11\) No breadcrumbs when searching or displaying form answers:
image


12\) Error on target change:
image

13\) I have a test form with 6 targets, when I submit it 0 of them are created.

Here is the form export (had to save it as .json.txt as json file are forbidden it seems).

[export_formcreator_20220119_1521.json.txt](https://github.com/pluginsGLPI/formcreator/files/7897870/export_formcreator_20220119_1521.json.txt)

14\) I have this warning in my logs:
```
[2022-01-19 15:01:53] glpiphplog.NOTICE:   *** PHP Notice (8): Undefined index: is_captcha_enabled in /home/adrie/localhost/gmaster2/plugins/formcreator/front/form_profile.form.php at line 48  
```

btry:
> 1) The visibility condition dropdown is too large, it doesn't seem to fit the input:

Reduced; but the width is not consistent everywhere. There is a problem with design of questions. This form is not converted into Twig and I integrated the twig version of conditions rendering in the old code. 

A full technical redesign of this part is needed because of several reasons (incoming devs, a communautary problem report, migration to twig, constant addition of options making this part had to maintain, bug prone). 

I want to fix that in a distinct PR to have a chance to submit a readable and understandable code design. This PR is too big and spreads on to many topics (fixes unrelated ti twig conversion) now.

> Also, if you have a scroll-able page (a big question form like the checkbox question type for example), the options are disconnected from the input:

should  be fixed with a speecific PR, see above comment

> 2) When creating a form, "All langages" is displayed differently when selected vs the value in the dropdown

fixed in GLPI, was a RTL problem. Typo fixed.

> 3) I'm not sure what happened but the first form I tried to create didn't had the correct default values.

fixed. Was a problem of data stored in session, not erased when a updating form properties, not changing the form object itself. see PluginFormcreatorForm::post_updateItem:() 

Maybe a better solution should be found as it is currently hacky.


> 4) The validation input still "moves around" depending on the selected value

out of topic of this PR : this tab is not converted into Twig yet. Also not very harmful, other things have priority

AdrienClairembault (approved):
OK for now, it's hard to keep track of everything with so many comments, commits and changes.
I'll test everything again later (next beta release maybe ?)

btry:
I'm still fixing last issues; i'll update the previous comment. I('ll tell you when tie PR is ready for a new review

PR #2549, opened 2022-01-26
Bad field name

btry:
Fix a non existing field name

very likely be an error in 0a0fcad17a288f8ad33ab2179cf4031af1364363

### DIFF
--- a/inc/abstracttarget.class.php
+++ b/inc/abstracttarget.class.php
@@ -1551,7 +1551,7 @@ SCRIPT;
       echo '<td width="25%">';
 
       // COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP
-      $display = $this->fields['commonitil_validation_answer_user'] == self::COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP ? "" : "display: none";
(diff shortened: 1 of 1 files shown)


inc/abstracttarget.class.php
```diff
@@ -1551,7 +1551,7 @@ protected function showValidationSettings($rand) {
       echo '<td width="25%">';
 
       // COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP
-      $display = $this->fields['commonitil_validation_answer_user'] == self::COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP ? "" : "display: none";
+      $display = $this->fields['commonitil_validation_question'] == self::COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP ? "" : "display: none";
```
AdrienClairembault, line 1554: Seems like it should be `commonitil_validation_rule`.
```suggestion
      $display = $this->fields['commonitil_validation_rule'] == self::COMMONITIL_VALIDATION_RULE_SPECIFIC_USER_OR_GROUP ? "" : "display: none";
```

Also can you fix this one too ?
image

AdrienClairembault (changes requested):

AdrienClairembault (approved):

PR #2591, opened 2022-02-14
html rendering in targets

### LINKED ISSUE #2568: Incorrect display of some tickets via form.
Currently under GLPI 10.0.0 with the version of formcreator Alpha.3. Some tickets appear this way when you open them, although the preview in the list is correct :

image
image

While some other tickets are displayed correctly either in preview or when opening the ticket.
In addition, when creating a new ticket, the display error also appears.

btry:
Closes #2568

By the way, this PR ensures the ITIL objects are created with the same scaping scheme as GLPI does internally.

### DIFF
--- a/inc/field/actorfield.class.php
+++ b/inc/field/actorfield.class.php
@@ -35,7 +35,6 @@ namespace GlpiPlugin\Formcreator\Field;
 use PluginFormcreatorAbstractField;
 use Html;
 use User;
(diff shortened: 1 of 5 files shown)


AdrienClairembault (approved):
Good fix, I've experienced this issue on my side too but wasn't able to reproduce it consistently

PR #2597, opened 2022-02-16
Count forms must take into account visibility

btry:
When counting forms to show or not the menu entry in "Assistance", the new columns is_visible is not taken into account.

is_visible allows the form to be accessible from assistance (or service catalog), whereas is_visible = 0 means the form is only acessible from its direct URL.


AdrienClairembault (approved):
Good find, I've missed this part.

btry:
I'll prepare a distinct PR to mutualize the mandatory conditions in the 2 main places.

PR #2605, opened 2022-02-17
TargetTicket: support 'Last valid answer' option for location

AdrienClairembault:
### Changes description

Add 'Last valid answer' option for location:

image

### DIFF
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -32,6 +32,7 @@
 use GlpiPlugin\Formcreator\Exception\ImportFailureException;
 use GlpiPlugin\Formcreator\Exception\ExportFailureException;
 use Glpi\Application\View\TemplateRenderer;
+use GlpiPlugin\Formcreator\Field\DropdownField;
 
 if (!defined('GLPI_ROOT')) {
    die("Sorry. You can't access this file directly");
@@ -633,6 +634,7 @@ class PluginFormcreatorTargetTicket extends PluginFormcreatorAbstractTarget
                case self::LOCATION_RULE_SPECIFIC:
                   $input['location_question'] = $input['_location_specific'];
                   break;
+               case self::LOCATION_RULE_LAST_ANSWER:
                default:
                   $input['location_question'] = '0';
             }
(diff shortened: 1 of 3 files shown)


inc/targetticket.class.php
```diff
@@ -971,6 +973,57 @@ protected function setTargetLocation($data, $formanswer) {
          case self::LOCATION_RULE_SPECIFIC:
             $location = $this->fields['location_question'];
             break;
+         case self::LOCATION_RULE_LAST_ANSWER:
+            $form_id = $formanswer->fields['id'];
```
btry, line 977:
```suggestion
            $formAnswerId = $formanswer->getID();
```

inc/targetticket.class.php
```diff
@@ -971,6 +973,57 @@ protected function setTargetLocation($data, $formanswer) {
...
+               'JOIN' => [
+                  PluginFormcreatorQuestion::getTable() . ' AS question' => [
+                     'ON' => [
+                        'answer' => 'plugin_formcreator_questions_id',
+                        'question' => 'id',
+                     ]
+                  ]
+               ],
+               'WHERE' => [
+                  'answer.plugin_formcreator_formanswers_id' => $form_id,
```
btry, line 993:
```suggestion
                  'answer.plugin_formcreator_formanswers_id' => $formAnswerId,
```

inc/targetticket.class.php
```diff
@@ -971,6 +973,57 @@ protected function setTargetLocation($data, $formanswer) {
...
+                  'question.fieldtype'                       => "dropdown",
```
btry, line 994:
```suggestion
                  'question.fieldtype'                       => 'dropdown',
```

inc/targetticket.class.php
```diff
@@ -971,6 +973,57 @@ protected function setTargetLocation($data, $formanswer) {
...
+               }
+
+               // Skip if question is not visible
+               if (!$formanswer->isFieldVisible($answer['plugin_formcreator_questions_id'])) {
+                  continue;
+               }
+
+               // Found a valid answer, stop here
+               $location = $answer['answer'];
+               break;
```
btry, line 1024: The feature should use the lastr valid answer. If we break; the code will select the 1st valid answer. I think this break should  be removed.

btry (commented):
I believe I found an inconsistency between the description of the feature and what the code actually does. From my point of view, the code selects the 1st valid answer, not the last one.

inc/targetticket.class.php
AdrienClairembault: It would if we started from the first answer, but the trick here is that we are starting from the end of the form:

image
image

inc/targetticket.class.php
AdrienClairembault: Done in 875fd405 + a few other occurrences I found from similar settings.

inc/targetticket.class.php
AdrienClairembault: Not sure about this one, GLPI coding standards allow both syntaxes.
I think it's best to use the same "core" coding standards for all plugins or we'll end up with 10 different standards very quickly 😅

btry: GLPI coding standards allow both syntaxes, then using single quotes is better as there is a very slight performance gain. Repeated by thousands of strings, there is probably a measurable benefit. Green coding.

btry:
Replaced by #2608 which contains nit tests

PR #2619, opened 2022-02-24
fix ticket content with html

btry:


btry:
does not solve the issue.

PR #2624, opened 2022-03-02
Update menu icons to Tabler-Icons

AdrienClairembault:
Convert existing fa-icons to ti-icons to match changes done in GLPI core.

### DIFF
--- a/inc/common.class.php
+++ b/inc/common.class.php
@@ -923,13 +923,14 @@ JAVASCRIPT;
          $newMenu['seek_assistance'] = [
             'default' => Plugin::getWebDir('formcreator', false) . '/front/wizard.php',
             'title'   => __('Seek assistance', 'formcreator'),
-            'icon'    => 'fa fa-paper-plane',
+            'icon'    => 'fa-fw ti ti-send',
          ];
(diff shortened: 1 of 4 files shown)


tests/code-coverage/clover.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
btry, line 1: this file should not be in the repo

btry (changes requested):

AdrienClairembault:
Is this the one were you told me that this hook isn't needed since we can already redefine menus with a core hook ?

I was worried of what would happen if the plugins redefining the menus were loaded before formcreator.
It would maybe be best to have some kind of code in the plugins that force them to be loaded after formcreator (thus no need for this menu hook).

I'll remove the menu part and keep the new icons if that's ok with you, since most GLPI menus have been migrated to them.

btry (dismissed):

AdrienClairembault:
Rebase + force-pushing to check if tests are ok  now

btry (approved):

PR #2622, opened 2022-03-01
twig question edition

btry:
linked to https://github.com/glpi-project/glpi/pull/10806

Contains commits to fix the unit tests witrh the font awesome dropdown.

PR #2644, opened 2022-03-16
Option show all or default forms

btry:
based on #2622

new configuration to show only default forms on virtual root form category

image

test build below
[glpi-formcreator-2.13.0-alpha.3-dev.zip](https://github.com/pluginsGLPI/formcreator/files/8266892/glpi-formcreator-2.13.0-alpha.3-dev.zip)

### DIFF
--- a/install/upgrade_to_2.13.php
+++ b/install/upgrade_to_2.13.php
@@ -38,6 +38,7 @@ class PluginFormcreatorUpgradeTo2_13 {
    public function upgrade(Migration $migration) {
       $this->migration = $migration;
       $this->migrateEntityConfig();
+      $this->addDefaultFormListMode();
       $this->fixRootEntityConfig();
       $this->migrateFkToUnsignedInt();
       $this->addFormAnswerTitle();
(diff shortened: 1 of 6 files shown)


btry:
This pr should solve both

AdrienClairembault (changes requested):
The fix seems to be working fine.

I now see why you added the new configuration option, it was indeed something we discussed in the possibles UX improvement, so OK for me on this part.

So the fix is working and the new option is justified but there is still a functional issue regarding the old "Default form in service catalog" option.

image

In the "Only default forms" mode, it only acts a second 'active' option for forms without categories.

Indeed if a form is without category, it needs this option + active to be displayed.
If either one if these option is missing, the form is never displayed.

With that in mind I think the "Default form in service catalog" option could be deleted without issue as it no longer serve any purposes (a "default" form is a form without categories, no need for another option).

Removing this option would make the "Only default forms" label in the entity configuration no longer relevant (since there is no notion of "default form" anymore).
To keep things as clear as possibles for users, I think something like the following labels would make sense:
"Forms shown when no category is selected" -> "All forms" / "Forms in the root category"

For even more clarity, the empty value label in the "Form Category" dropdown could indicate that the form will be placed in the top level category (I suggest "Root category" as we already call the top entity the "Root entity", keeping the same wording).

image


What are your thoughts on this ?

install/upgrade_to_2.13.php
```diff
@@ -46,7 +47,7 @@ public function upgrade(Migration $migration) {
       $this->addTargetValidationSetting();
       $this->addFormVisibility();
       $this->addDashboardVisibility();
-      $this->addRequestSourceSeting();
+      $this->addRequestSourceSetting();
```
btry, line 50: typo

PR #2646, opened 2022-03-21
Manage documents in initial notifications

### LINKED ISSUE #2614: Attached files missing in mail notification for new tickets
Hi,

Long time ago, we discussed about missing filenames and wrong counter in email notifications for new tickets created through Formcreator ( https://forum.glpi-project.org/viewtopic.php?id=251773 ). I was pretty sure to had opened a bug report for this, but it seems that I didn't...

In the meantime I created some messy patches as workaround, but now I try to find what happens... It seems that PluginFormcreatorTargetTicket->save() function is looking for document name in the content, but there is only an 'Attached document' instead ( foreach (PluginFormcreatorCommon::getDocumentsFromTag($data['content']) as $document) )

kabassanov:
### Changes description

Manage documents in initial notifications. Before this changes, target tickets are created and notifications are sent before documents get attached to the ticket. So initial notifications by e-mail don't contain references to them.
With these changes, documents are attached during ticket creation following standard creation procedure. It also works for forms requiring validation. 

The disadvantage of the proposed solution is that it manages document insertion twice (but using standard glpi duplication detection, files are not duplicated).

 !-- issues related (for reference or to be closed) and/or links of discuss -->

Closes #2614

### DIFF
--- a/inc/formanswer.class.php
+++ b/inc/formanswer.class.php
@@ -1572,4 +1572,41 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
       return PluginFormcreatorFields::isVisible($this->questionFields[$id]->getQuestion(), $this->questionFields);
    }
+
+//KKK
+   public function getFileName($questionId,$index) {
+      $doc = new Document();
+      $doc->getFromDB($this->questionFields[$questionId]->getDocumentsForTarget()[$index]);
+      return $doc->fields['filepath'];
+   }
+
+   public function getFileFields() : Array {
+      $filefields=[];
+      $form = $this->getForm();
+      foreach ($this->getQuestionFields($form) as $questionId => $field) {
+	 $question = $field->getQuestion();
+	 if ($question->fields['fieldtype'] === 'file') 
+	    $filefields[] = $questionId;
+      }
+
+      return $filefields;
+   }
+
+   public function getFileProperties() : Array {
+      $form = $this->getForm();
+      foreach ($this->getFileFields() as $questionId) {
+	 foreach ($this->getQuestionFields($form)[$questionId]->getDocumentsForTarget() as $key => $documentId){
+	    $doc = new Document();
+	    $doc->getFromDB($documentId);
+	    $formcreator_field[]= $doc->fields['filename'];
+	    $tag_formcreator_field[]= $doc->fields['tag'];
+	 }
+	 $formcreator_field_array[$questionId] = $formcreator_field;
+	 $tag_formcreator_field_array[$questionId] = $tag_formcreator_field;
+      }
+
+      return [ "_filename" => $formcreator_field_array,
+	       "_tag_filename" => $tag_formcreator_field_array];
+   }
+
 }
--- a/inc/targetticket.class.php
+++ b/inc/targetticket.class.php
@@ -816,6 +816,32 @@ class PluginFormcreatorTargetTicket extends PluginFormcreatorAbstractTarget
          $data['_tag_content'][] = $document['tag'];
       }
 
+   //KKK
+   $saved_documents = $formanswer->getFileProperties();
+
+   if ($saved_documents) {
+      foreach ($formanswer->getFileFields() as $questionId) {
+	 $data["_filename"]=array_merge($data["_filename"],$saved_documents["_filename"][$questionId]);
+	 $data["_tag_filename"]=array_merge($data["_tag_filename"],$saved_documents["_tag_filename"][$questionId]);
+
+	 foreach ( $saved_documents["_filename"][$questionId] as $key => $filename) {
+	    $uploaded_filename=$formanswer->getFileName($questionId,$key);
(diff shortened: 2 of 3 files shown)


btry:
Additional note : if it works well for t ickets, this contrub should be applied also in other taget types (changes and problems)

inc/formanswer.class.php
```diff
@@ -1572,4 +1572,41 @@ public function isFieldVisible(int $id) : bool {
...
+	 if ($question->fields['fieldtype'] === 'file') 
+	    $filefields[] = $questionId;
+      }
+
+      return $filefields;
+   }
+
+   public function getFileProperties() : Array {
+      $form = $this->getForm();
+      foreach ($this->getFileFields() as $questionId) {
```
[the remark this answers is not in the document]
kabassanov: Yes, it is possible to do this. But getFileFields will still be used in inc/targetticket.class.php as iterator for fields of file type.

inc/targetticket.class.php
```diff
@@ -816,6 +816,32 @@ public function save(PluginFormcreatorFormAnswer $formanswer) {
...
+   } else {
+      foreach ($formanswer->getFileFields() as $questionId) {
+	 $data["_filename"]=array_merge($data["_filename"],$formanswer->input["_formcreator_field_".$questionId]);
+	 $data["_prefix_filename"]=array_merge($data["_prefix_filename"],$formanswer->input["_prefix_formcreator_field_".$questionId]);
+	 $data["_tag_filename"]=array_merge($data["_tag_filename"],$formanswer->input["_tag_formcreator_field_".$questionId]);
+	 foreach ( $formanswer->input["_formcreator_field_".$questionId] as $key => $filename) {
+	    $uploaded_filename=$formanswer->getFileName($questionId,$key);
+	    copy(GLPI_DOC_DIR . '/' . $uploaded_filename, GLPI_TMP_DIR . '/' . $filename);
+	 }
+      }
```
btry, line 842: Hi

It seems we never enter this part of code.

When Formcreator calls save() method, we always have a form answer available. I did some tests and as I expected I could not enter here. I think it is dead code. Can you tell in which case this part is used ?

Note: This is the last point to solve before finishing my port for Formcretor 2.13. If you have a GLPI 10 for tests, you may want to test the version I prepared.

kabassanov: Let me check... 
The first thing I can see is that my getFileProperties function never returns NULL. I have to change this in order to avoid php warnings when no file is attached.
Next, it seems that you are right. As far as I remember, I created the first part of the if statement in order to manage forms requiring validation. It seems that the test is also true for immediate ticket creation as files are already in the DB. But this way we are loosing eventual _prefix_filename field from the input...

Actually, I don't have GLPI 10 instance...

btry: wait before changing your PR. Have a look instead on my version for 2.13 : I reworked the code of this method (and it moved in an abstract class).

>
As far as I remember, I created the first part of the if statement in order to manage forms requiring validation. 

I tested this case, and even when the form is under validation we don't enter in the part I pointed. This is normal : the form answer is created before  the validation step.

Then in all cases we loose the _prefix_ and this might be a problem to have the documents in the notifications. Notification is the last thing I must to check on my side

kabassanov: With the PR code, I have documents in the notifications as I copy files back to the temp folder. That is probably the reason why missing _prefix_filename is not really an issue... Need confirmation of "prefix" role in the original code...
The else statement is not correct and not needed as if getFileProperties returns an empty array we don't have to fill documents fields.

Function remake of getFileProperties should be better:

   public function getFileProperties() : Array {

      $formcreator_field=[];
      $tag_formcreator_field=[];
      $formcreator_field_array=[];
      $tag_formcreator_field_array=[];

      $form = $this->getForm();

      foreach ($this->getFileFields() as $questionId) {

         foreach ($this->getQuestionFields($form)[$questionId]->getDocumentsForTarget() as $key => $documentId){
            $doc = new Document();
            $doc->getFromDB($documentId);
            $formcreator_field[]= $doc->fields['filename'];
            $tag_formcreator_field[]= $doc->fields['tag'];
         }
         if ($formcreator_field && $tag_formcreator_field) {
            $formcreator_field_array[$questionId] = $formcreator_field;
            $tag_formcreator_field_array[$questionId] = $tag_formcreator_field;
         }
      }

      if ($formcreator_field_array && $tag_formcreator_field_array) {
         return [ "_filename" => $formcreator_field_array,
                  "_tag_filename" => $tag_formcreator_field_array];
      } else {
        return [];
      }
   }

Not sure to have time to check 2.13 today...

btry: Hi 

FYI : the version I built from your PR for 2.13 has been merged.

btry:
Hi

 Having no feedback about #2648, I assume you did not had time to validate. No major problem reported by beta testers so far.

PR #2647, opened 2022-03-21
Prepare to support non Itil targets

btry:


AdrienClairembault (changes requested):
Seems OK overall, only one issue here:

image

`getItem_Item` may be null.

You can either use the solution from #2590:
image

Or move `getItem_Item` and part of this function logic to `AbstractITILTarget`

btry:
getGenertadTargets() is used to find the aggregated status if a form answer. 

I think I would pick the 1st solution : the case will happen when the target does not actually create a persisted object in DB.

PR #2648, opened 2022-03-22
WIP : Add documents to targets before creating them

btry:
This is a port of #2646 to version 2.13, with additional support for changes and problems.

It must solve an old limitation in Formcreator impacting the notifications of a generated ticket : the notification does not contains the documents of the generated ticket. The cause was the documents are added to the ticket after its creation.

This PR adds the documents at creation time.


btry:
replaced by #2654

AdrienClairembault (approved):

PR #2651, opened 2022-03-23
assistance menu - deprecation error

btry:
Deprecation error 

```
<PHP Deprecated: Automatic conversion of false to array is deprecated in C:\inetpub\wwwroot\glpi\plugins\formcreator\inc\formlist.class.php on line 51

```

### DIFF
--- a/inc/formlist.class.php
+++ b/inc/formlist.class.php
@@ -46,12 +46,11 @@ class PluginFormcreatorFormList extends CommonGLPI
       return _n('Form', 'Forms', $nb, 'formcreator');
    }
 
-   public static function getMenuContent() {
-      $menu = parent::getMenuContent();
-      $menu['title'] = static::getTypeName(Session::getPluralNumber());
-      $menu['page'] = PluginFormcreatorFormList::getSearchURL(false);
-      $menu['icon'] = 'fas fa-edit';
+   public static function canView() {
+      return (PluginFormcreatorForm::countAvailableForm() > 0);
+   }
 
-      return $menu;
+   public static function getIcon() {
+      return 'fas fa-edit';
    }
 }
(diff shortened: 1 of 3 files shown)


inc/formlist.class.php
```diff
@@ -1,57 +0,0 @@
...
-    *
-    * @param number $nb Number of item(s)
-    * @return string Itemtype name
-    */
-   public static function getTypeName($nb = 0) {
-      return _n('Form', 'Forms', $nb, 'formcreator');
-   }
-
-   public static function getMenuContent() {
-      $menu = parent::getMenuContent();
```
btry, line 50: getMenuContent() may return false

inc/formlist.class.php
```diff
@@ -46,12 +46,11 @@ public static function getTypeName($nb = 0) {
...
    }
 
-   public static function getMenuContent() {
-      $menu = parent::getMenuContent();
-      $menu['title'] = static::getTypeName(Session::getPluralNumber());
-      $menu['page'] = PluginFormcreatorFormList::getSearchURL(false);
-      $menu['icon'] = 'fas fa-edit';
+   public static function canView() {
+      return (PluginFormcreatorForm::countAvailableForm() > 0);
+   }
```
btry, line 51: In CommonGLPI::getMenuContent() : if  the itemtype is not inherited from CommonDBTM, calls canView() if it exists.

The menu entry must show if there is at least 1 form available for the user. This is done in PluginFormcreatorForm::countAvailableForm.

The menu is created like does CommonGLPI::getMenuContent() (except $menu['links']['search'] which does not matters). Then we can simply drop getMenuContent() here and let the parent implementation do its job

inc/formlist.class.php
```diff
@@ -46,12 +46,11 @@ public static function getTypeName($nb = 0) {
...
 
-      return $menu;
+   public static function getIcon() {
+      return 'fas fa-edit';
    }
```
btry, line 55: Properly define an icon.

PR #2658, opened 2022-03-26
WIP only full group by

btry:
While debugging the sync issues SQL query, I encountered several problems like the following

```
Expression #2 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'glpi-master.glpi_tickets.name' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
```

This PR fixes those errors


btry:
No solution found. Might need a complete redesign if the query, maybe by splitting it into several separate queries. As this problem impacts mostly debugging context, let's postpone this task

PR #2659, opened 2022-03-28
Simplify sync issue

btry:
remove useless nested subquery.

Might have a slight positive impact on performance. 

Note : there are unit tests dedicated to this query, then this change is safe, as long it passes tthem.

### DIFF
--- a/inc/issue.class.php
+++ b/inc/issue.class.php
@@ -182,18 +182,15 @@ class PluginFormcreatorIssue extends CommonDBTM {
             ],
             [
                'TABLE' => new QuerySubQuery([
-                  'SELECT' => '*',
-                  'FROM' => new QuerySubQuery([
-                     'SELECT' => ['users_id', $ticketFk],
-                     'DISTINCT' => true,
-                     'FROM'  => $ticketUserTable,
-                     'WHERE' => [
-                        'type' => CommonITILActor::REQUESTER,
-                     ],
-                     'ORDER' => ['id ASC'],
-                  ], 'inner_glpi_tickets_users'),
-                  'GROUPBY' => 'tickets_id'
-               ], 'glpi_tickets_users'),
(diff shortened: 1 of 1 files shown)


inc/issue.class.php
```diff
@@ -162,8 +162,8 @@ public static function getSyncIssuesRequest() : AbstractQuery {
                'entities_id                              as entities_id'
             ],
             new QueryExpression('0                       as is_recursive'),
-            new QueryExpression("COALESCE($ticketUserTable.users_id, 0) as requester_id"),
-            new QueryExpression("IF(`$ticketValidationTable`.`users_id_validate` IS NULL, 0, `$ticketValidationTable`.`users_id_validate`)  as users_id_validator"),
+            new QueryExpression("COALESCE(`$ticketUserTable`.`users_id`, 0) as `requester_id`"),
+            new QueryExpression("COALESCE(`$ticketValidationTable`.`users_id_validate`, 0) as `users_id_validator`"),
```
btry, line 166: - escape table / column names
- use coalesce instead of IF ELSE, (more readable expression)

inc/issue.class.php
```diff
@@ -182,18 +182,15 @@ public static function getSyncIssuesRequest() : AbstractQuery {
...
-               ], 'glpi_tickets_users'),
+                  'SELECT' => ['users_id', $ticketFk],
+                  'DISTINCT' => true,
+                  'FROM'  => $ticketUserTable,
+                  'WHERE' => [
+                     'type' => CommonITILActor::REQUESTER,
+                  ],
+                  'GROUPBY' => 'tickets_id',
+                  'ORDER' => ['id ASC']
+               ], $ticketUserTable),
```
btry, line 193: while searching to make the query full_group_by compliant, I found it is possible to remove 1 nesting of SELECT.

AdrienClairembault (approved):

PR #2665, opened 2022-03-30
Fix form display

AdrienClairembault:
### Changes description

Main issue was that https://github.com/pluginsGLPI/formcreator/commit/152ba276e17605308bde409d47cdd3c45d9c88d4 changed the `getRenderedHtml` method signature but the associated code in templates was not updated.
Also a few variables were not referenced properly (`$this->fields` instead of `$this->question->fields` and `domain` instead of `options.domain`).

There is a few bonus fixes regarding other issues I encountered while testing this feature.

Full changes:
* Fix form display
* Set input type to "number" for integer questions
* Fix targets (`getItem_Item` is not always defined)
* Add simple test to ensure displaying a form doesn't trigger any errors

### DIFF
--- a/templates/pages/userform.html.twig
+++ b/templates/pages/userform.html.twig
@@ -47,12 +47,12 @@
     {% set formName = 'plugin_formcreator_form' %}
 	<form name="{{ formName }}" method="post" role="form" enctype="multipart/form-data" class="plugin_formcreator_form" action={{ item.getFormURL() }} id="plugin_formcreator_form" data-itemtype="PluginFormcreatorForm" data-id="{{ item.fields['id'] }}">
 		<h1 class='form-title'>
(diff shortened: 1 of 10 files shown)


templates/pages/userform.html.twig
```diff
@@ -91,7 +91,7 @@
                         {% endif %}
                         {% if not options.public or question.getSubField().isPublicFormCompatible() %}
                             {% set sessionData = session('formcreator') %}
-                            {{ question.getRenderedHtml(domain, true, sessionData.data)|raw }}
```
AdrienClairembault, line 94: Is the `null` part correct or is there an actual `FormAnswer` object that should be referenced here ?

btry:
obsoleted by https://github.com/pluginsGLPI/formcreator/pull/2670

PR #2666, opened 2022-03-30
refactor(formanswer): supports generated targets without relation table

btry:
optimize a method

### DIFF
--- a/inc/abstractitiltarget.class.php
+++ b/inc/abstractitiltarget.class.php
@@ -2241,8 +2241,8 @@ SCRIPT;
       return $data;
    }
 
-   public function isObjectTarget(): bool {
-      return true;
+   public static function getTargetType(): int {
+      return self::TARGET_TYPE_OBJECT;
    }
 
    /**
--- a/inc/abstractitiltarget.class.php
+++ b/inc/abstractitiltarget.class.php
@@ -96,9 +96,9 @@ PluginFormcreatorTranslatableInterface
     * Get an instance object for the relation between the target itemtype
     * and an object of any itemtype
     *
-    * @return CommonDBTM
+    * @return CommonDBRelation
     */
-   abstract public function getItem_Item();
+   abstract public static function getItem_Item(): CommonDBRelation;
 
    /**
     * Get the class name of the target itemtype's template class
@@ -2239,5 +2239,50 @@ SCRIPT;
       }
 
       return $data;
+   }
+
+   public function isObjectTarget(): bool {
+      return true;
+   }
+
+   /**
+    * Find generated targets for this target and the given form answer
+    *
+    * @param PluginFormcreatorFormAnswer $formAnswer
+    * @return array
+    */
+   public static function findForFormAnswer(PluginFormcreatorFormAnswer $formAnswer): array {
+      global $DB;
+
+      $targets = [];
+      $relationType = static::getItem_Item();
+      $relationTable = $relationType::getTable();
+      $generatedType = static::getTargetItemtypeName();
+      $generatedTypeTable = $generatedType::getTable();
+      $fk = $generatedType::getForeignKeyField();
+      $iterator = $DB->request([
+         'SELECT' => ["$generatedTypeTable.*"],
+         'FROM' => $generatedTypeTable,
+         'INNER JOIN' => [
+            $relationTable => [
+               'FKEY' => [
+                  $generatedTypeTable => 'id',
+                  $relationTable => $fk,
+               ],
+            ],
+         ],
+         'WHERE' => [
+            "$relationTable.itemtype" => PluginFormcreatorFormAnswer::getType(),
+            "$relationTable.items_id" => $formAnswer->getID(),
+         ],
+      ]);
+      foreach ($iterator as $row) {
+         /** @var $item CommonDBTM */
+         $item = new $generatedType();
+         $item->getFromResultSet($row);
+         $targets[] = $item;
+      }
+
+      return $targets;
    }
 }
--- a/inc/formanswer.class.php
+++ b/inc/formanswer.class.php
@@ -1849,7 +1849,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
       }
       /** @var PluginFormcreatorTargetInterface $targetType */
       $this->targetList = [];
-      foreach (PluginFormcreatorForm::getTargetTypes() as $targetType) {
+      foreach ($itemtypes as $targetType) {
          $targetItem = new $targetType();
          $generatedType = $targetItem->getTargetItemtypeName();
          $relationType = $targetItem->getItem_Item();
@@ -1880,10 +1880,6 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
             $item = new $generatedType();
             $item->getFromResultSet($row);
             $this->targetList[] = clone $item;
-            // skip not wanted itemtypes
-            if (!in_array($targetType, $itemtypes)) {
-               continue;
-            }
             $targets[] = $item;
          }
       }
--- a/inc/formanswer.class.php
+++ b/inc/formanswer.class.php
@@ -863,7 +863,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
             // Generate the target
             $generatedTarget = $targetObject->save($this);
-            if ($targetObject->isObjectTarget()) {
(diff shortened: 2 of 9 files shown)


AdrienClairembault (dismissed):

inc/abstractitiltarget.class.php
```diff
@@ -96,9 +96,9 @@ abstract protected function getItem_Supplier();
     * Get an instance object for the relation between the target itemtype
     * and an object of any itemtype
     *
-    * @return CommonDBTM
+    * @return CommonDBRelation
```
btry, line 99: more precise type hinting

inc/abstractitiltarget.class.php
```diff
@@ -96,9 +96,9 @@ abstract protected function getItem_Supplier();
...
     */
-   abstract public function getItem_Item();
+   abstract public static function getItem_Item(): CommonDBRelation;
```
btry, line 101: switching to static method to avoif creating an object for this call which does not uses $this

inc/abstractitiltarget.class.php
```diff
@@ -2240,4 +2240,43 @@ protected function prepareUploadedFiles(array $data, $formanswer): array {
 
       return $data;
    }
+
+   public function isObjectTarget(): bool {
+      return true;
+   }
```
btry, line 2246: provide info about the task of this target : it generates an object

inc/abstractitiltarget.class.php
```diff
@@ -2240,4 +2240,49 @@ protected function prepareUploadedFiles(array $data, $formanswer): array {
...
+
+   /**
+    * Find generated targets for this target and the given form answer
+    *
+    * @param PluginFormcreatorFormAnswer $formAnswer
+    * @return array
+    */
+   public static function findForFormAnswer(PluginFormcreatorFormAnswer $formAnswer): array {
```
btry, line 2254: this method comes from the move of the SQL query from PluginFormcreatorFormAnswer::getGeneratedTargets()

inc/formanswer.class.php
```diff
@@ -862,13 +863,16 @@ public function generateTarget(): bool {
...
-               $success = false;
-               break;
+            if ($targetObject->isObjectTarget()) {
+               if ($generatedTarget === null) {
+                  // If the target generates an object but none generated
+                  $success = false;
+               } else {
+                  $this->targetList[] = $generatedTarget;
+                  // Map [itemtype of the target] [item ID of the target] = ID of the generated target
+                  $generatedTargets->addTarget($targetObject, $generatedTarget);
```
btry, line 873: populate targetList only for generataed objects, and feed the composite tickets linker only in this case.

inc/formanswer.class.php
```diff
@@ -44,6 +44,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    /**
     * Generated targets after creation of a form answer
+    * Populated only after creation of a FormAnswer in DB with add() method
```
btry, line 47: this property is used only to completesmart HTTP redirection after ceating a formanswer from a form. See front/form.form.php

inc/formanswer.class.php
```diff
@@ -1828,15 +1832,11 @@ public function  getCurrentApprovers(): ?array {
     * get all generated targets by the form answer
     * populates the generated targets associated to the instance
     *
-    * Assumes that the generated target has a relation with the form answer
```
btry, line 1831: No longer true

inc/formanswer.class.php
```diff
@@ -1848,44 +1848,8 @@ public function getGeneratedTargets($itemtypes = []): array {
          $itemtypes = array_intersect(PluginFormcreatorForm::getTargetTypes(), $itemtypes);
       }
       /** @var PluginFormcreatorTargetInterface $targetType */
-      $this->targetList = [];
-      foreach (PluginFormcreatorForm::getTargetTypes() as $targetType) {
```
btry, line 1852: moved in AbstractItilTarget

inc/targetinterface.class.php
```diff
@@ -37,5 +37,27 @@ interface PluginFormcreatorTargetInterface
...
+    */
+   public function isObjectTarget(): bool;
+
+   /**
+    * Find targets linked to a form answer
+    *
+    * @param PluginFormcreatorFormAnswer $formAnswer
+    * @return array
+    */
+   public static function findForFormAnswer(PluginFormcreatorFormAnswer $formAnswer): array;
```
btry, line 62: 3 new mandatory methods

tests/3-unit/PluginFormcreatorFormAnswer.php
```diff
@@ -699,23 +699,16 @@ public function testGetGeneratedTargets() {
       $generatedTargets = $instance->targetList;
 
       $output = $instance->getGeneratedTargets();
-      // Check the targetsList is unchanged
       $this->array($instance->targetList)->hasSize(count($generatedTargets));
       $this->array($output)->hasSize(3);
 
       $output = $instance->getGeneratedTargets([PluginFormcreatorTargetTicket::getType()]);
-      // Check the targetsList is unchanged
-      $this->array($instance->targetList)->hasSize(3);
```
btry, line 708: getGeneratedObjects does no longer populate targetList : useless.

inc/targetinterface.class.php
```diff
@@ -37,5 +37,27 @@ interface PluginFormcreatorTargetInterface
...
+    * @return boolean true if the target does an action
+    */
+   public function isActionTarget(): bool;
+
+   /**
+    * Tell if this target generates an object and does not does an action
+    *
+    * @return boolean true if the target generates an object
+    */
+   public function isObjectTarget(): bool;
```
AdrienClairembault, line 54: The comments seems to indicate that `isActionTarget` and `isObjectTarget` are exclusive.
Wouldn't a single method `getTargetType` returning something like `AbstractTarget::TYPE_ACTION` or `AbstractTarget::TYPE_OBJECT` be better ?

`isActionTarget` and `isObjectTarget` could stays as helper methods in the base abstract class but only `getTargetType` would be required by the interface.

inc/targetinterface.class.php
```diff
@@ -37,5 +37,27 @@ interface PluginFormcreatorTargetInterface
...
 
-   public function getTargetItemtypeName();
+   public static function getTargetItemtypeName();
+
+   /**
+    * Tell if this target does an action and does not generate an object
+    *
+    * @return boolean true if the target does an action
+    */
+   public function isActionTarget(): bool;
```
AdrienClairembault, line 47: You set multiple similar functions as static in this PR, maybe this one should be static too ?

inc/targetinterface.class.php
```diff
@@ -37,5 +37,27 @@ interface PluginFormcreatorTargetInterface
...
+
+   /**
+    * Tell if this target generates an object and does not does an action
```
AdrienClairembault, line 50:
```suggestion
    * Tell if this target generates an object and does not do an action
```

AdrienClairembault (changes requested):
Looks good overall expected for a few minor changes (see comments above).

If I get it correctly, targets like `EmailTarget` from advanced forms does not count as generating "objects".
Even if `EmailTarget` create a `SentEmail` object, since this object is only used to confirm that the action was run properly, it is not expected to be returned by `findForFormAnswer`.

inc/targetinterface.class.php
btry: > The comments seems to indicate that isActionTarget and isObjectTarget are exclusive

Yes I consider they are exclusive. Actually creating a ticket is already an action by itself, but it creates an object on the DB. Your SentEmail is also an object, then EmailTarget generates an object. 

> The comments seems to indicate that isActionTarget and isObjectTarget are exclusive.
Wouldn't a single method getTargetType returning something like AbstractTarget::TYPE_ACTION or AbstractTarget::TYPE_OBJECT be better ?
>
> **isActionTarget** and **isObjectTarget** could stays as helper methods in the base abstract class but only getTargetType would be required by the interface.

If I switch to a single method returning a constant, then I wwould drop **isActionTarget** and **isObjectTarget**  from this PR.

inc/targetinterface.class.php
btry: I agree, but not in this PR. And I focus on releasing an Alpha 4. I wanted to release it last week already.

AdrienClairembault (approved):

PR #2669, opened 2022-03-31
fix(form): bad var to speecify the translation domain

btry:


AdrienClairembault:
Why not merge https://github.com/pluginsGLPI/formcreator/pull/2665 directly ? Tests are greens.

PR #2676, opened 2022-04-03
bump(2.13.0-alpha.4): 2.13.0-alpha.4

flegastelois:
### Changes description

To be able to download alpha.4 from marketplace, bump xml and setup.


btry:
Hi,

I did this change in master branch already. It may be worth to check why it was not taken into account.

Merged. If you have a moment, let me know if te plugin has been published properly.

PR #2683, opened 2022-04-05
fix(fieldsfield): instanciation error when fields plugin disabled

btry:
We cannot instanciate classes from Fields plugin in constructor. This throws a fatal error.

I added a lazy loading of the instance of  the class, and provided a way to edit a fields question anyway.

### DIFF
--- a/inc/field/fieldsfield.class.php
+++ b/inc/field/fieldsfield.class.php
@@ -38,15 +38,11 @@ use PluginFormcreatorForm;
 use Dropdown;
 use DbUtils;
 use Plugin;
-use Ticket;
-use Change;
-use Problem;
 use Session;
 use Html;
 use OperatingSystem;
 use PluginFieldsDropdown;
 use PluginFieldsField;
-use PluginFormcreatorQuestion;
 use Toolbox;
 use User;
 
@@ -59,19 +55,20 @@ class FieldsField extends PluginFormcreatorAbstractField
    public $field = null;
 
    /**
+    * Get the additional field object lined to this formcreator field
     *
-    * @param array $question PluginFormcreatorQuestion instance
+    * @return null|PluginFieldsField
     */
-   public function __construct(PluginFormcreatorQuestion $question) {
-      $this->question  = $question;
-      if (isset($this->question->fields['values'])) {
+   public function getField(): ?PluginFieldsField {
+      if ($this->field === null && isset($this->question->fields['values'])) {
          $decodedValues = json_decode($this->question->fields['values'], JSON_OBJECT_AS_ARRAY);
          $field_name = $decodedValues['dropdown_fields_field'] ?? '';
          $fieldObj = new PluginFieldsField();
          if ($fieldObj->getFromDBByCrit(['name' => $field_name])) {
-            $this->field  = $fieldObj;
+            $this->field = $fieldObj;
          }
       }
+      return $this->field;
    }
 
    public function isPrerequisites(): bool {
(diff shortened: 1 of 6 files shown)


inc/field/fieldsfield.class.php
```diff
@@ -38,15 +38,11 @@
...
 use Plugin;
-use Ticket;
-use Change;
-use Problem;
 use Session;
 use Html;
 use OperatingSystem;
 use PluginFieldsDropdown;
 use PluginFieldsField;
-use PluginFormcreatorQuestion;
```
btry, line 49: those classes re not used, cleaning use statements

inc/field/fieldsfield.class.php
```diff
@@ -58,20 +54,16 @@ class FieldsField extends PluginFormcreatorAbstractField
...
+      if ($this->field === null && isset($this->question->fields['values'])) {
          $decodedValues = json_decode($this->question->fields['values'], JSON_OBJECT_AS_ARRAY);
          $field_name = $decodedValues['dropdown_fields_field'] ?? '';
          $fieldObj = new PluginFieldsField();
          if ($fieldObj->getFromDBByCrit(['name' => $field_name])) {
-            $this->field  = $fieldObj;
+            $this->field = $fieldObj;
          }
       }
+      return $this->field;
```
btry, line 66: still a risk to return null here; either we should throw an exception, either we should alow returning null and handle it after each call of this method.

inc/field/fieldsfield.class.php
```diff
@@ -152,16 +144,23 @@ public function getBlocks(): array {
...
 
    public function showForm(array $options): void {
+      if (!\Plugin::isPluginActive('fields')) {
+         $template = '@formcreator/field/undefinedfield.html.twig';
+         TemplateRenderer::getInstance()->display($template, [
+            'item' => $this->question,
+            'params' => $options,
+         ]);
+         return;
+      }
```
btry, line 154: Ensure we can show the modal for the question and edit it even when Fields is disabled

inc/field/fieldsfield.class.php
```diff
@@ -152,16 +149,23 @@ public function getBlocks(): array {
...
```
btry, line 159: show an template for an undefined field as the plugin Additional Fields is not available and some HTML inputs cannot be rendered

inc/field/fieldsfield.class.php
```diff
@@ -152,16 +149,24 @@ public function getBlocks(): array {
...
    public function showForm(array $options): void {
+      if (!\Plugin::isPluginActive('fields')) {
+         $options['error'] = __('Warning: Additional Fields plugin is disabled or missing', 'formcreator');
+         $template = '@formcreator/field/undefinedfield.html.twig';
+         TemplateRenderer::getInstance()->display($template, [
+            'item' => $this->question,
+            'params' => $options,
+         ]);
+         return;
+      }
```
btry, line 160: show an error message if ADditioinal fields is disabled, while allowing to edit the question

inc/field/tagfield.class.php
```diff
@@ -46,6 +46,16 @@ public function isPrerequisites(): bool {
...
    public function showForm(array $options): void {
+      if (!\Plugin::isPluginActive('tag')) {
+         $options['error'] = __('Warning: Tag plugin is disabled or missing', 'formcreator');
+         $template = '@formcreator/field/undefinedfield.html.twig';
+         TemplateRenderer::getInstance()->display($template, [
+            'item' => $this->question,
+            'params' => $options,
+         ]);
+         return;
+      }
```
btry, line 57: for tag question type: provide also an error message

inc/fields.class.php
```diff
@@ -107,7 +108,7 @@ public static function getNames() {
       // Get localized names of field types
       foreach (PluginFormcreatorFields::getClasses() as $field_type => $classname) {
          $classname = self::getFieldClassname($field_type);
-         if ($classname == PluginFormcreatorTagField::class && !$plugin->isActivated('tag')) {
+         if ($classname == UndefinedField::class) {
```
btry, line 111: exclude undefined field from available types. Leave tags field available.

We may want to maintain tag and fields types unavailable, if we add a new method in PluginFormcreatorFirldInterface, with a slight difference with isPrerequisite(). one for dependencies out of the plugin, the other for some kind of validation of question settings.

PR #2724, opened 2022-04-26
feat: rights management

Rom1-B:
### Changes description

Addition of a tab for setting FC rights on profiles. This makes it possible to assign rights to forms without having rights to entities.

### Checklist

### References

!23662

### DIFF
--- a/front/form_language.form.php
+++ b/front/form_language.form.php
@@ -31,7 +31,7 @@
 
 include ('../../../inc/includes.php');
 
-Session::checkRight('entity', UPDATE);
+Session::checkRight(PluginFormcreatorForm::$rightname, UPDATE);
 
 // Check if plugin is activated...
 if (!(new Plugin())->isActivated('formcreator')) {
--- a/inc/profile.class.php
+++ b/inc/profile.class.php
@@ -31,37 +31,37 @@
 
 class PluginFormcreatorProfile extends Profile {
 
-    function getTabNameForItem(CommonGLPI $item, $withtemplate = 0) {
-        return self::createTabEntry(PluginFormcreatorForm::getTypeName(Session::getPluralNumber()));
-    }
+   function getTabNameForItem(CommonGLPI $item, $withtemplate = 0) {
+      return self::createTabEntry(PluginFormcreatorForm::getTypeName(Session::getPluralNumber()));
+   }
 
-    static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0) {
-        $formcreatorprofile = new self();
-        $formcreatorprofile->showForm($item->getID());
-        return true;
-    }
+   static function displayTabContentForItem(CommonGLPI $item, $tabnum = 1, $withtemplate = 0) {
+      $formcreatorprofile = new self();
+      $formcreatorprofile->showForm($item->getID());
+      return true;
+   }
 
-    function showForm($ID, $options = []) {
-        if (!self::canView()) {
-            return false;
-        }
+   function showForm($ID, $options = []) {
+      if (!self::canView()) {
+         return false;
+      }
 
-        echo "<div class='spaced'>";
-        $profile = new Profile();
-        $profile->getFromDB($ID);
-        echo "<form method='post' action='".$profile->getFormURL()."'>";
+      echo "<div class='spaced'>";
+      $profile = new Profile();
+      $profile->getFromDB($ID);
+      echo "<form method='post' action='".$profile->getFormURL()."'>";
 
-        $rights = [['itemtype'  => 'PluginFormcreatorForm',
-                    'label'     => PluginFormcreatorForm::getTypeName(Session::getPluralNumber()),
-                    'field'     => 'plugin_formcreator']];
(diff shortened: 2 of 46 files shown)


inc/profile.class.php
```diff
@@ -0,0 +1,67 @@
...
+        if (!self::canView()) {
+            return false;
+        }
+
+        echo "<div class='spaced'>";
+        $profile = new Profile();
+        $profile->getFromDB($ID);
+        echo "<form method='post' action='".$profile->getFormURL()."'>";
+
+        $rights = [['itemtype'  => 'PluginFormcreatorForm',
```
btry, line 54:
```suggestion
        $rights = [['itemtype'  => PluginFormcreatorForm::getType(),
```

setup.php
```diff
@@ -374,7 +374,7 @@ function plugin_formcreator_hook(): void {
    $PLUGIN_HOOKS[Hooks::REDEFINE_MENUS]['formcreator'] = [PluginFormcreatorCommon::class, 'hookRedefineMenu'];
 
    // Config page
-   if (Session::haveRight('entity', UPDATE)) {
+   if (Session::haveRight(PluginFormcreatorForm::$rightname, UPDATE)) {
```
btry, line 377: With your feature I think that we should check if the user has at least one right among CREATE, READ, UPDATE, DELETE, PURGE and maybe the import right if we implement it (see comments above)

btry (changes requested):

btry:
The plugin grant righs to the profile of the current user in the installation process. The current user is usally a super admin.

front/form_language.form.php
```diff
@@ -31,7 +31,7 @@
 
 include ('../../../inc/includes.php');
 
-Session::checkRight('entity', UPDATE);
+Session::checkRight(PluginFormcreatorForm::$rightname, UPDATE);
```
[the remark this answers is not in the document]
Rom1-B: I think this would add unnecessary complexity.
If the profile only has the right to update and not the right to create, does that mean that it can modify translations, but not add them?

install/install.php
```diff
@@ -99,6 +99,31 @@ public function install(Migration $migration, $args = []): bool {
...
+         if ($rights[\Entity::$rightname] >= UPDATE) {
+            \ProfileRight::updateProfileRights($profile['id'], [
+               \PluginFormcreatorForm::$rightname => READ + UPDATE + CREATE + DELETE + PURGE,
+            ]);
+         } else {
+            \ProfileRight::updateProfileRights($profile['id'], [
+               \PluginFormcreatorForm::$rightname => READ,
+            ]);
+         }
+      }
```
btry, line 126:
```suggestion
      foreach ($profiles as $profile) {
         $rights = ProfileRight::getProfileRights(
            $profile['id'],
            [
               Entity::$rightname,
               PluginFormcreatorForm::$rightname,
            ]
         );
         $right = READ;
         if ($rights[Entity::$rightname] & (UPDATE + CREATE + DELETE + PURGE)) {
            $right = READ + UPDATE + CREATE + DELETE + PURGE;
         }
         ProfileRight::updateProfileRights($profile['id'], [
            PluginFormcreatorForm::$rightname => $right,
         ]);
      }
```

btry: This suggestion 
- reduces sligthly the code, 
- drops canonical class names (if GLPI core classes move in a namespace this will be risky and painful to update everywhere. It is easier to simply add a **use** statement at the beginning of the file)
- use bitwise masks expressions instead of comparing bit fields as integer values

install/install.php
```diff
@@ -100,6 +100,31 @@ public function install(Migration $migration, $args = []): bool {
       $task = new CronTask();
       PluginFormcreatorIssue::cronSyncIssues($task);
 
+      // Add rights
```
btry, line 103: The whole logic after this line to set rights in profiles should be in a dedicated method.

Also, rights must be added when upgrading the plugin. I don't think it is possible to factorize your code to add rights below; you will have to copy it in upgrade_to_2_14.php

btry (changes requested):
Here are some changes and suggestions  to do. See also my PR to add unit tests on the added rights.

inc/profile.class.php
```diff
@@ -0,0 +1,67 @@
...
+      }
+
+      echo "<div class='spaced'>";
+      $profile = new Profile();
+      $profile->getFromDB($ID);
+      echo "<form method='post' action='".$profile->getFormURL()."'>";
+
+      $rights = [['itemtype'  => PluginFormcreatorForm::getType(),
+                  'label'     => PluginFormcreatorForm::getTypeName(Session::getPluralNumber()),
+                  'field'     => 'plugin_formcreator']];
```
btry, line 56: I found a bug : the right uses the old name. Fix below.

Bould you by the way rebase your branch to get all updates done on develop ? 

```suggestion
                  'field'     => PluginFormcreatorForm::$rightname]];
```

btry (changes requested):
a bug has been found. I'm doing some tests with the suggested change.

btry (approved):

PR #2753, opened 2022-05-05
Add option to hide searchissue

stonebuzz:
### Changes description

New option to display or not search Issue (juste toggle the switch)

image

image

### Checklist

### References

Closes #N/A


inc/entityconfig.class.php
```diff
@@ -315,6 +326,22 @@ public function showFormForEntity(Entity $entity) {
       }
       echo '</td></tr>';
 
+      // Search issue visibility
+      $elements = self::getEnumSearchIssueVisibility();
+      if ($entityId == 0) {
+         unset($elements[self::CONFIG_PARENT]);
+      }
+      echo "<tr class='tab_bg_1'>";
+      echo "<td>".__('Search issue')."</td>";
```
btry, line 335:
```suggestion
      echo "<td>".__('Search issue', 'formcreator')."</td>";
```

btry (changes requested):
A suggestion + need to have same defaults for install and upgrade.

We need to consider to have this feature avaialble in GLPI core, as I don't like much to use JS to hide things in the UI.

stonebuzz:
> We need to consider to have this feature avaialble in GLPI core, as I don't like much to use JS to hide things in the UI.

done =)

install/upgrade_to_2.13.php
```diff
@@ -48,6 +48,14 @@ public function upgrade(Migration $migration) {
...
       $this->addFormVisibility();
       $this->addRequestSourceSetting();
+      $this->addEntityOption();
+   }
+
+   public function addEntityOption() {
+      $table = 'glpi_plugin_formcreator_entityconfigs';
+
+      $this->migration->addField($table, 'is_search_issue_visible', 'integer', ['after' => 'is_header_visible', 'value' => '-2']);
+      $this->migration->addPostQuery("UPDATE `glpi_plugin_formcreator_entityconfigs` SET `is_search_issue_visible`= 1 WHERE `entities_id` = 0");
```
btry, line 58: these 2 lines must run if the field does not already exists. Currently, running the upgrade twice (after a failed attempt or a forced upgrade) may overwrite the setting for the root entity.

btry (changes requested):

PR #2793, opened 2022-05-20
Do not give access to tickets list

btry:
In Formcreator 2.13 direct acess to tickets list is no longer useful. If a user types the URL to this page the plugin should redirect to the issues list page.


btry:
obsoleted by #2796

PR #2858, opened 2022-06-28
Do no show refused tickets in the "New" category

AdrienClairembault:
Internal ref: https://support.teclib.com/front/ticket.form.php?id=24344

Refused forms shouldn't be shown when searching for "New" tickets

image


AdrienClairembault:
Refused items are already shown in the "resolved" tab.
An item can't be both new and resolved.

AdrienClairembault:
Refused items are already shown in the "resolved" tab.
An item can't be both new and resolved.

PR #2907, opened 2022-08-01
Embedded image upload broken

btry:
Image uploads in textarea is hacky and broke, ~probably due to switching to unsigned int IDs and FK.~

This PR solved the problem by converting uploads as documents immediately, contrary to previous implemntation. To handle generation of tickets the uploaded stream is re-used to simulate an upload, like it is currently implemented for questions of type file.

When showing a form answer from simplified interface, documents are not accessible. We need to provide itemtype and items_id identifiers to an object linked to the document, and viewable by the curent user.
(see https://github.com/glpi-project/glpi/pull/10945 ). Formcreator cannot rely on GLPI to build such URL for now.

Needs an improvement to avoid a duplicate of its code in the plugin.


inc/field/actorfield.class.php
```diff
@@ -306,7 +307,10 @@ public function prepareQuestionInputForSave($input) {
       }
 
       $this->value = $parsed;
-      $input['default_values'] = $this->serializeValue();
+      $input['default_values'] = '';
+      if ($this->value !== null && $this->value != '') {
+         $input['default_value'] = json_encode($this->value);
+      }
```
btry, line 313: avoid calling serializeValue as we don't have a form answer to provide

PR #2909, opened 2022-08-01
Ensure uploads are kept after form reload or rebuild from answers

btry:


btry:
It appears to be useless in PR #2907.

PR #2914, opened 2022-08-03
fix: cannot delete a ticket from service catalog

btry:
internal ref !24537 topic 5


btry:
Validated by the bug reporter

PR #2915, opened 2022-08-03
Show title in navigation header

btry:
In service catalog, tickets are displayes without their title. To ensure consistent interface, formanswers and issues should also show their titme in the navigation header.

This PR relies on https://github.com/glpi-project/glpi/pull/12372

The screenshot below show a status dics or circle. This is no longer true because of a change in the PR for GLPI, allowing to not implement getStatusIcon()

### a ticket in service catalog
image

### a form answer in service catalog

image

### DIFF
--- a/inc/formanswer.class.php
+++ b/inc/formanswer.class.php
@@ -41,6 +41,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
    public $dohistory  = true;
    public $usenotepad = true;
    public $usenotepadrights = true;
+   protected static $showTitleInNavigationHeader = true;
 
    /**
     * Generated targets after creation of a form answer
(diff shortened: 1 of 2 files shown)


inc/formanswer.class.php
```diff
@@ -522,10 +523,12 @@ public function showForm($ID, $options = []) {
...
-      echo "<h1 class='form-title'>";
-      echo $this->fields['name'] . "&nbsp;";
-      echo '<i class="pointer print_button fas fa-print" title="' . __("Print this form", 'formcreator') . '" onclick="window.print();"></i>';
-      echo '</h1>';
+      if (version_compare(GLPI_VERSION, '10.0.3') < 0) {
+         echo "<h1 class='form-title'>";
+         echo $this->fields['name'] . "&nbsp;";
+         echo '<i class="pointer print_button fas fa-print" title="' . __("Print this form", 'formcreator') . '" onclick="window.print();"></i>';
+         echo '</h1>';
+      }
```
btry, line 531: fallback if the requirements from GLPI are not met.

wc55:
Hi

I don't know if this is the correct place for this, but would it be possible to include the ticket number next to the title, just like it is on the technician's screen?

PR #2922, opened 2022-08-11
Fx possible inconsistency in issues table

btry:


btry:
already  handled in an other upgrade task...

PR #2923, opened 2022-08-11
feature(form,section,question): remove ability to get all objects of form from API

btry:
Feature no longer used, and causes an infinite call loop when loading a form object from the API


inc/field/dropdownfield.class.php
```diff
@@ -154,13 +154,11 @@ public function buildParams($rand = null) {
             $decodedValues['entity_restrict'] = $decodedValues['entity_restrict'] ?? 2;
             switch ($decodedValues['entity_restrict']) {
                case self::ENTITY_RESTRICT_FORM:
-                  $form = PluginFormcreatorForm::getByItem($this->getQuestion());
```
btry, line 157: $form is already defined line 115

inc/field/dropdownfield.class.php
```diff
@@ -154,13 +154,11 @@ public function buildParams($rand = null) {
...
                   $currentEntity = $form->fields['entities_id'];
                   $ancestorEntities = getAncestorsOf(Entity::getTable(), $currentEntity);
                   break;
 
                case self::ENTITY_RESTRICT_BOTH:
-                  $form = PluginFormcreatorForm::getByItem($this->getQuestion());
```
btry, line 163: $form is already defined line 115

AdrienClairembault (approved):

PR #2961, opened 2022-09-05
fix saving of tag question and specifics in target ticket/change

VladoTTX:
### Changes description
Target ticket and change does not check proper input values on tag questions it should be `_tag_questions` and `_tag_specifics` 

https://github.com/pluginsGLPI/formcreator/blob/8413e0af88a473ef5cfc1b17c4c1d5585d9c2381/inc/targetticket.class.php#L687-L698

But looks like TargetProblem does not have this check implemented for some reason at all..

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/targetchange.class.php
+++ b/inc/targetchange.class.php
@@ -581,12 +581,12 @@ class PluginFormcreatorTargetChange extends PluginFormcreatorAbstractItilTarget
          }
 
          if (Plugin::isPluginActive('tag')) {
-            if (isset($input['tag_questions'])) {
+            if (isset($input['_tag_questions'])) {
                $input['tag_questions'] = (!empty($input['_tag_questions']))
                                           ? implode(',', $input['_tag_questions'])
                                           : '';
             }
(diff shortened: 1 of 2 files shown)


btry:
Hi

I noticed in a recent development that a multiselect empty field does not sends its value when it is empty. Instead it sends a post variable names by the multiselect name prefixed with an underscore.

Therefore, I'm not sure your fix will  or break the feature.
 By the way if the problems can

VladoTTX:
I meant that in TargetProblem you don't have this check on existence of `_tag_` key implemented at all:
https://github.com/pluginsGLPI/formcreator/blob/8413e0af88a473ef5cfc1b17c4c1d5585d9c2381/inc/targetproblem.class.php#L310-L317

Instead you copy directly values from underscore values `_tag_*`

btry (approved):

btry:
I checked the POST data of the form. Your PR is OK. I also added a factorization to enable tags in problems. For some reason, the feature was incomplete or damaged.

PR #2978, opened 2022-09-14
Restrict glpi object items by filter

btry:
### Changes description

internal ref: 23137

Requires : 
- https://github.com/glpi-project/glpi/pull/12637
- https://github.com/btry/glpi/commit/f3d66d65796262dbe0e8a82e7bde78c0fa0786c8
- https://github.com/glpi-project/glpi/pull/12615
- https://github.com/glpi-project/glpi/commit/40400792cc48e25fe6015f97e8a983154998cc5a (for backport to 10.0/bugfixes)
- https://github.com/glpi-project/glpi/pull/13289 (amends the above commit)
- https://github.com/pluginsGLPI/fields/pull/593

### Checklist

Please check if your PR fulfills the following specifications:

- [ ] Tests for the changes have been added
- [ ] Docs have been added/updated

### References

Closes #N/A

PR #2995, opened 2022-09-22
Tabs in question designer, question conditions moved in tab

btry:
### Requires 
-  GLPI 10.0.3
- https://github.com/glpi-project/glpi/pull/12740

Pick from #2978 of tab feature for question modals

Move question conditions in a new tab

Closes #N/A


btry:
obsolete; develop branch is back to version 2.14, because of big features which broke the move from inc/ to src/

PR #2999, opened 2022-09-23
Email addresses were ignored

btry:
### Changes description

Email addresses (without user matching a GLPI internal account) were ignored

internal ref 24999


inc/abstractitiltarget.class.php
```diff
@@ -2182,9 +2182,6 @@ public function cleanActors(array $data): array {
                   $cleaned = [];
                   $cleaned_notif = [];
                   foreach ($data["$actorType"] as $key => $actor) {
-                     if ($actor == 0) {
-                        continue;
-                     }
```
btry, line 2187: if $actor == 0 then we have an accountless email address

cedric-anne (commented):
Seems OK.

PR #3003, opened 2022-09-27
Backslash escaping, broken form in preview mode

btry:
internal ref 24982


btry:
Obsoleted by fixes + unit tests

PR #3010, opened 2022-10-03
fix(form): submit once

AdrienClairembault:
### Changes description

Prevent submitting a form multiple times (client side).

It is mostly done to prevent user errors (`e.g.` clicking the button multiple time while waiting for a response).
If you want to also prevent this on the server side to avoid potential abuse, feel free to suggest another PR ;)

### References

!25029

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -1381,6 +1381,10 @@ var plugin_formcreator = new function() {
       var form     = document.querySelector('form[data-itemtype]');
       var data     = new FormData(form);
       data.append('submit_formcreator', '');
+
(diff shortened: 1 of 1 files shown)


btry (approved):
I guess the error fixed here is related to a CSRF token used twice by clicking again the submit button. This solution should be sufficient.

btry:
Reverted on favor of this commit: abed861018c4a04b3385c43223a2554eaf4dcaf7

GLPI already has this feature, and impleemtns a spinner. This should resolve users stressed by delays after clicking the submit button.

PR #3031, opened 2022-10-12
Fix FAQ list

AdrienClairembault:
### Changes description

After fixing the FAQ tree in https://github.com/glpi-project/glpi/pull/12922, I've noticed that it was still incorrect in formcreator.
If you have at least one visible article then all categories will be shown (even if these categories have no visible articles themselves).

This seems caused by an incorrect SQL query:

image

The column that is supposed to count the visible items in a category actually count the number of visible articles in the whole FAQ (as you can see in the `WHERE` clause - no restriction on current category).

If I compare the code to the one used in the core (there is two very similar functions - maybe formcreator should reuse the core code here instead of rewriting it).
Formcreator `inc.knowbase.class.php`:
image

Core  `Src/Knowbase.php`:
image

You can see that the core code have the correct `WHERE` clause that limit the scope to the current category, while the formcreator code seems to have an useless `JOIN` instead (that must be removed for the `WHERE` to work as it target the same table).

After copying the core code, the SQL request work as expected:

image

### References

!25026
https://github.com/glpi-project/glpi/pull/12922

### DIFF
--- a/inc/knowbase.class.php
+++ b/inc/knowbase.class.php
@@ -112,13 +112,12 @@ class PluginFormcreatorKnowbase {
                            KnowbaseItem_KnowbaseItemCategory::getTable() => KnowbaseItem::getForeignKeyField(),
                      ],
                   ],
-                  KnowbaseItemCategory::getTable() => [
-                     'FKEY' => [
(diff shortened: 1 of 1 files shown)


AdrienClairembault:
I've added another commit to use the same INNER join as the core aswell (not sure if it changes anything, but let's keep the same behavior in both core and plugin to be safe).

btry (approved):

PR #3093, opened 2022-11-23
More hooks

AdrienClairembault:
### Changes description

Some additionals hooks needed for advancedforms.

Two very simple hooks:
* `formcreator_question_description` triggered when displaying a question's description 
* `formcreator_question_default_value` triggered when displaying a question's default value

And a more complex hook `formcreator_actors_type` that allow to define custom actors.

This hooks expect a class name, which must implement the new `PluginFormcreatorPluginTarget` interface.
You'll find an example here: https://gist.github.com/AdrienClairembault/89346d3e545260651f8375a772fc06e6, which allow to select an actor specified in a question from a previous form of the same workflow.

image

The only impact on existing code is a rewrite of the `changeActor` javascript which handle hiding/showing the custom html forms for each actors types.

This javascript code now rely on data attributes instead of specific instructions, which was needed to allow plugins using the `formcreator_actors_type` hook to have their specific html forms displayed.

Also it seems that the same `changeActor` was written twice, I've deleted the second occurrence.


inc/plugintarget.class.php
```diff
@@ -0,0 +1,47 @@
...
+ * @link      https://pluginsglpi.github.io/formcreator/
+ * @link      http://plugins.glpi-project.org/#/plugin/formcreator
+ * ---------------------------------------------------------------------
+ */
+
+if (!defined('GLPI_ROOT')) {
+   die("Sorry. You can't access this file directly");
+}
+
+interface PluginFormcreatorPluginTarget
```
btry, line 36: Class name should be PluginFormcreatorPluginTargetInterface, because it is an interface and ither interfaces use this suffix.

btry (changes requested):

btry (approved):

PR #3144, opened 2023-01-10
fix (form): Invalid event target and debouncing

AdrienClairembault:
### Changes description

Form with multiple textarea have terrible performance, resulting easily in thousands of AJAX request when editing a single field.

Two issues :

1\) Wrong event target

`input` event watcher are not assigned to the correct textarea, they all go to the first one.
So if you have 10 textarea, the first one will watch for `input` 10 times (thus triggering 10 ajax request to `showfields.php` per change) and the others will not be watched at all.

2\) No debouncing

Typing a few words into a textarea may generate close to one hundred `input` events (and thus an hundred ajax request to `showfields.php`).
We solve this with debouncing, grouping multiple events into one if they are less than 400ms apart.

### References

!25974

### DIFF
--- a/js/scripts.js
+++ b/js/scripts.js
@@ -865,6 +865,8 @@ var plugin_formcreator = new function() {
       });
    };
 
+   this.showFieldsDebounced = _.debounce(this.showFields, 400, false);
+
    // === SECTIONS ===
 
    this.deleteSection = function (item) {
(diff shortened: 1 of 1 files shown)


btry:
I think there is a risk that a condition to show / hide a question or even the submit button will run too late. If a user sends the form before the timeout actually recompute the visibilities, we may have problems.

I would rather consider to move visibility computation client side, as it looks safer to me (but the work is bigger).

AdrienClairembault:
I think you are overthinking it, debouncing is fairly commonly used on forms and wont be an issue 99.9% of the time.
In fact, users can probably already submit the form while the ajax request is pending, which has not been a big problem this far.
I'm not even sure more than 0.1% of the plugin's users have a condition on the content of a _textarea_, it's pretty niche.

And if it comes to that, nothing prevent us from disabling submission until the `showFields.php` call resolve (even though I wont be implementing this in this pull request as I don't want to spend too much time for an edge case with my current limited availabilities - feel free to implement it yourself if you think it's important).

Respectfully, planning a full rewrite of the fields visibility computations on the client side for something that can be effectively fixed in less than 10 lines of code seems like a big waste of man hours to me.

btry:
OK, then let's apply the patch; and let's keep in mind we can tune the delay (400ms is quite high then it probably provides some margin).

btry (approved):

btry:
Cherry-picked in support/2.13.0. Commit 4f482f48b08c2224e0f13a1a5e778d9067bf8b1b

PR #3151, opened 2023-01-18
fix: Support new core option for assets visibility

AdrienClairembault:
### Changes description

Support new option added in GLPI 10.1: [13832](https://github.com/glpi-project/glpi/pull/13832)

I assume formcreator 2.14 would be for GLPI 10.1 only but I'm not 100% sure (can't remember what's been decided on that).
If that's not the case then I'll add some methodExist check to not break things with GLPI 10.0.


btry:
Hi

Formcreator targets GLPI 10.0 and 10.1. this an requirement, and I will try to maintain it as much as possible.

btry (approved):

PR #3169, opened 2023-02-07
Encoding problem when image inserted in content

btry:
### Changes description

When an image is embedded in a textarea question, a ticket generated and using this question shows HTML tags.

### Checklist

### References

internal ref 26534

### DIFF
--- a/inc/field/textareafield.class.php
+++ b/inc/field/textareafield.class.php
@@ -210,7 +210,6 @@ class TextareaField extends TextField
          );
          $input[$key] = $this->value; // Restore the text because we don't want image converted into A + IMG tags
          // $this->value = $input[$key];
-         $this->value = Sanitizer::unsanitize($this->value);
          foreach ($input['_tag'] as $docKey => $tag) {
             $newTag = $this->uploads['dedup'][$tag];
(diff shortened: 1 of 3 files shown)


inc/field/textareafield.class.php
```diff
@@ -210,7 +210,6 @@ public function serializeValue(PluginFormcreatorFormAnswer $formanswer): string
          );
          $input[$key] = $this->value; // Restore the text because we don't want image converted into A + IMG tags
          // $this->value = $input[$key];
-         $this->value = Sanitizer::unsanitize($this->value);
```
btry, line 213: value is unsanitized here, then no need to unsanitize it.

tests/3-unit/GlpiPlugin/Formcreator/Field/TextareaField.php
```diff
@@ -168,7 +168,7 @@ public function testDeserializeValue($question, $input, $expected) {
 
       $instance->parseAnswerValues($input);
       $instance->deserializeValue($input[$key]);
-      $output = $instance->getValueForTargetText('', false);
+      $output = $instance->getValueForTargetText('', true);
```
btry, line 171: the change in the expected value reveals that this bool was wrongly set to false.

PR #3181, opened 2023-02-17
Compute priority

btry:
### Changes description

internal ref !26713

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/abstractitiltarget.class.php
+++ b/inc/abstractitiltarget.class.php
@@ -428,7 +428,6 @@ PluginFormcreatorTranslatableInterface
       }
       if (!is_null($urgency) && $urgency != 0) {
          $data['urgency'] = $urgency;
-         $data['priority'] = CommonITILObject::computePriority($data['urgency'], $data['impact']);;
       }
 
       return $data;
--- a/inc/abstractitiltarget.class.php
+++ b/inc/abstractitiltarget.class.php
@@ -434,6 +434,18 @@ PluginFormcreatorTranslatableInterface
    }
 
    /**
+    * Calculate priority from the predefined urgency and impact
+    *
+    * @param array $data
+    * @return array the updated data for the ITIL object
+    */
+   protected function setTargetPriority(array $data): array {
+      $data['priority'] = CommonITILObject::computePriority($data['urgency'], $data['impact']);
+
+      return $data;
+   }
+
+   /**
     * find all actors and prepare data for the ticket being created
     */
    protected function prepareActors(PluginFormcreatorForm $form, PluginFormcreatorFormAnswer $formanswer) {
--- a/inc/targetchange.class.php
+++ b/inc/targetchange.class.php
@@ -675,7 +675,7 @@ class PluginFormcreatorTargetChange extends PluginFormcreatorAbstractItilTarget
          $data[$changeField] = $formanswer->parseTags($data[$changeField], $this, $changeField == 'content');
       }
 
-      $data['_users_id_recipient']   = $_SESSION['glpiID'];
+      $data['_users_id_recipient']   = $formanswer->fields['requester_id'];
 
       $this->prepareActors($form, $formanswer);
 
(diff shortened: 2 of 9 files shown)


tests/src/CommonTestCase.php
```diff
@@ -268,6 +268,8 @@ protected function getTargetTicket($input = []) {
 
       $targetTicket = new \PluginFormcreatorTargetTicket();
       $targetTicket->add($input);
+      $this->boolean($targetTicket->isNewItem())->isFalse();
+      $targetTicket->getFromDB($targetTicket->getID());
```
btry, line 272: reload the object on memory to ensure all fields are in sync with DB.

inc/targetchange.class.php
```diff
@@ -675,7 +675,7 @@ public function save(PluginFormcreatorFormAnswer $formanswer): ?CommonDBTM {
          $data[$changeField] = $formanswer->parseTags($data[$changeField], $this, $changeField == 'content');
       }
 
-      $data['_users_id_recipient']   = $_SESSION['glpiID'];
+      $data['_users_id_recipient']   = $formanswer->fields['requester_id'];
```
btry, line 678: Shall to be in sync with PluginFormcreatorTargetTicket implementation.

inc/targetproblem.class.php
```diff
@@ -194,7 +194,7 @@ public function save(PluginFormcreatorFormAnswer $formanswer): ?CommonDBTM {
          $data[$problemFields] = $formanswer->parseTags($data[$problemFields], $this, $problemFields == 'content');
       }
 
-      $data['_users_id_recipient']   = $_SESSION['glpiID'];
+      $data['_users_id_recipient']   = $formanswer->fields['requester_id'];
```
btry, line 197: Shall to be in sync with PluginFormcreatorTargetTicket implementation.

tests/3-unit/PluginFormcreatorTargetChange.php
```diff
@@ -30,9 +30,9 @@
  */
 
 namespace tests\units;
-use GlpiPlugin\Formcreator\Tests\CommonTargetTestCase;
+use GlpiPlugin\Formcreator\Tests\AbstractItilTargetTestCase;
 
-class PluginFormcreatorTargetChange extends CommonTargetTestCase {
+class PluginFormcreatorTargetChange extends AbstractItilTargetTestCase {
```
btry, line 35: to have one implementation of test for all ITIL targets

inc/abstractitiltarget.class.php
```diff
@@ -428,12 +428,23 @@ protected function setTargetUrgency($data, $formanswer) {
...
    }
 
+   /**
+    * Calculate priority from the predefined urgency and impact
+    *
+    * @param array $data
+    * @return array the updated data for the ITIL object
+    */
+   protected function setTargetPriority(array $data): array {
+      $data['priority'] = CommonITILObject::computePriority($data['urgency'], $data['impact']);
```
AdrienClairembault, line 443: You do not need to compute the priority, it will already be done in GLPI's core:

image

All we need to do is to unset `$input['priority']` to force its computation:

```suggestion
      // Remove default priority so it can be computed
      if (isset($data['urgency']) || isset($data['impact'])) {
         unset($data['priority']);
      }
```

AdrienClairembault: More details: the current version fail because you are calling `CommonITILObject::computePriority` which will never work because of the following code:

image

`static::MATRIX_FIELD` will always be empty from `CommonITILObject`:

image

So you would need to call it with this final item (`Ticket`, `Problem` or `Change`) or you wont be able to read the matrix.

With that being said, I think my original suggestion of letting the core handle it still the way to go.
If the process is already done in `prepareInputForAdd`, we don't need to duplicate it here.

btry:
replaced with #3194

PR #3215, opened 2023-03-24
version 2.13.5

btry:
### Checklist

### References

Closes #N/A


inc/command/cleanticketscommand.class.php
```diff
@@ -43,7 +43,7 @@ class CleanTicketsCommand extends Command
 {
    protected function configure() {
       $this
-         ->setName('glpi:plugins:formcreator:clean_tickets')
+         ->setName('plugins:formcreator:clean_tickets')
```
AdrienClairembault, line 46: Maybe the old name should be kept as an alias and only be removed on the next major version ? 
Same for the others commands names changes.

AdrienClairembault (approved):

PR #3241, opened 2023-04-04
Add Reminders

Unknoweb:
### Changes description

Add Reminders


btry:
Merged in support/2.13.0 with 6ed85cd366addb99b3f1a0d24ca62f18b7baaa69

this commit merges your PR with some changes
- no change in .po files (need to update locales in Transifex instead)
- change new lines from CRLF to LF (unix style)
- all changes in a single commit, with use of conventional commit

PR #3268, opened 2023-05-03
Show additional field value in fullform

btry:
### Changes description

When the tag FULLFORM is used, the additional field is not rendered, but the value is not available for self-service profile (ticket's right pane is unavailable). This PR fixes this.

### Checklist

### References

internal ref 27745


btry:
We won't merge : the requester may see the additional field it he has extended service catalog.

PR #3287, opened 2023-05-31
fix(buildParams): fix namespace

stonebuzz:
### Changes description

When testing version ```2.13.6```, the ```ID``` display in dropdowns (ticket type in my case) no longer works.

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/field/glpiselectfield.class.php
+++ b/inc/field/glpiselectfield.class.php
@@ -39,6 +39,7 @@ use Dropdown;
 use Entity;
 use CommonTreeDropdown;
 use CommonDBTM;
+use Ticket;
 use User;
 
 use GlpiPlugin\Formcreator\Exception\ComparisonException;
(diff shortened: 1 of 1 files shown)


inc/field/glpiselectfield.class.php
```diff
@@ -149,7 +149,7 @@ public function buildParams($rand = null) {
       if (isset($emptyItem->fields['otherserial'])) {
          $dparams['displaywith'][] = 'otherserial';
       }
-      if ($itemtype === Ticket::class && !array_search('id', $dparams['displaywith'])) {
+      if ($itemtype === \Ticket::class && !array_search('id', $dparams['displaywith'])) {
```
btry, line 152: It is better to use a ```use Ticket;``` statement in the top of the file

btry (changes requested):

inc/field/glpiselectfield.class.php
stonebuzz: done

PR #3326, opened 2023-07-13
fix(form): prevent duplicate form in list

stonebuzz:
### Changes description

Prevent duplicate form in central list

### Checklist

### References

Closes #N/A

### DIFF
--- a/inc/form.class.php
+++ b/inc/form.class.php
@@ -2001,7 +2001,6 @@ PluginFormcreatorTranslatableInterface
       }
 
       $form_query = PluginFormcreatorForm::getFormListQuery();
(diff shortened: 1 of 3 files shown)


inc/form.class.php
```diff
@@ -2001,6 +2001,7 @@ public function showForCentral() {
       }
 
       $form_query = PluginFormcreatorForm::getFormListQuery();
+      $form_query['DISTINCT'] = true;
```
btry, line 2004: Maybe this change should be in PluginFormcreatorForm::getFormListQuery() ?

stonebuzz: Indeed, I think it should be, in order to be a more global solution.

PR #3327, opened 2023-07-17
Fix search options

AdrienClairembault:
This condition break search options for forms with at least one target.
On the other hand, I'm not sure how this fix will impact forms without targets, can you confirm that it looks good to you @btry ?

!28799

### DIFF
--- a/inc/issue.class.php
+++ b/inc/issue.class.php
@@ -624,7 +624,7 @@ class PluginFormcreatorIssue extends CommonDBTM {
                   'table'                => PluginFormcreatorFormAnswer::getTable(),
                   'joinparams'           => [
                      'jointype'          => 'itemtype_item_revert',
-                     'specific_itemtype'  => PluginFormcreatorFormAnswer::class,
+                     'specific_itemtype'  => Ticket::class,
                   ]
                ],
             ],


AdrienClairembault:
After further testing, I wonder if search option 44 should maybe just be deleted ?
Search option 14 seems to be displaying the same data and work correctly :

image

Same for requester.

btry:
Hi

You're nearly right. 

Actually, SO 14 finds the technician for a issue having 1 and only 1 ticket, and SO 44 finds technicians for an issue having several tickets.

The task I began to solve the issue was to add in SO 44 the ability to solve technicians for both cases. With your finding, it would  be better to merge both into a single search option.

btry:
Additional note: Look into the search option 11. Usually a "before  join" key tells to join one table. This one is an example of how to join more tables. This is supported by the search engine. We need this pattern.

At some point in the SO 44, we need to follow 2 different path to find technicians for an issue with one ticket and to find technicians for issues having 2 or more tickets.

So far, for some reason, this solution does not works. The search engine tried to join with glpi_tickets with a wrong ON clause (glpi_tickets.items_id = ...) I'm suspecting a bug in the Search class.

PR #3346, opened 2023-08-03
Merge 2 search options

btry:
### Changes description

Search options 14 and 44 both provide the same information, but handle 2 different cases
- single ticket for an issue
- multiple tickets for an issue

The solution is hacky, must be improved if possible to remove the hardcoded table with MD5 hash from complex join (see Search.php in GLPI)

Alternative to #3327 

### Checklist

### References

Internal ref 28799


btry:
Too much complicated to do : would require support of unions subqueries joins in GLPI's Search class.

PR #3455, opened 2024-02-09
Fix identical status label in simplified dashboard

RomainLvr:
### Changes description

Update the STATUS_WAITING label to avoid an identical status label with PENDING. This confusion appears in `My requests for assistance` (Simplified dashboard) when GLPI is in French for example.
 
Before :
Capture d’écran du 2024-02-09 14-19-33
 
After : 
Capture d’écran du 2024-02-09 14-20-09

### Checklist

### References

Ticket [31528](https://support.teclib.com/front/ticket.form.php?id=31528)

### DIFF
--- a/inc/formanswer.class.php
+++ b/inc/formanswer.class.php
@@ -75,7 +75,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    public static function getStatuses() {
       return [
-         self::STATUS_WAITING  => __('Waiting', 'formcreator'),
+         self::STATUS_WAITING  => __('Waiting for approval'),
          self::STATUS_REFUSED  => __('Refused', 'formcreator'),
          self::STATUS_ACCEPTED => __('Accepted', 'formcreator'),
       ];


inc/formanswer.class.php
```diff
@@ -75,7 +75,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    public static function getStatuses() {
       return [
-         self::STATUS_WAITING  => __('Waiting', 'formcreator'),
+         self::STATUS_WAITING  => __('Waiting for approval'),
```
Rom1-B, line 78: In this case, the variable name is ambiguous

Rom1-B (changes requested):

inc/formanswer.class.php
RomainLvr: Fixed in [d53bd46](https://github.com/pluginsGLPI/formcreator/pull/3455/commits/d53bd4645b2ce3e43c66501a56aab5200bd333ce)

inc/formanswer.class.php
```diff
@@ -75,9 +76,10 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    public static function getStatuses() {
       return [
-         self::STATUS_WAITING  => __('Waiting', 'formcreator'),
-         self::STATUS_REFUSED  => __('Refused', 'formcreator'),
-         self::STATUS_ACCEPTED => __('Accepted', 'formcreator'),
+         self::WAITING_APPROVAL  => __('Waiting for approval'),
+         self::STATUS_WAITING    => __('Waiting', 'formcreator'),
+         self::STATUS_REFUSED    => __('Refused', 'formcreator'),
+         self::STATUS_ACCEPTED   => __('Accepted', 'formcreator'),
```
Rom1-B, line 82: 2 constants shouldn't have the same value.

I misspoke, I was thinking of renaming the 3 constants, as `STATUS_...` is not explicit enough.

```suggestion
         self::APPROVAL_WAITING    => __('Waiting for approval'),
         self::APPROVAL_REFUSED    => __('Refused', 'formcreator'),
         self::APPROVAL_ACCEPTED   => __('Accepted', 'formcreator'),
```

Rom1-B (changes requested):

inc/formanswer.class.php
```diff
@@ -57,6 +57,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    // Values choosen to not conflict with status of ticket constants
    // @see PluginFormcreatorIssue::getNewStatusArray
+   const WAITING_APPROVAL = 101;
    const STATUS_WAITING = 101;
    const STATUS_REFUSED = 102;
    const STATUS_ACCEPTED = 103;
```
Rom1-B, line 63: 2 constants shouldn't have the same value.

I misspoke, I was thinking of renaming the 3 constants, as STATUS_... is not explicit enough.

```suggestion
   const APPROVAL_WAITING = 101;
   const APPROVAL_REFUSED = 102;
   const APPROVAL_ACCEPTED = 103;
```

Rom1-B (changes requested):

inc/formanswer.class.php
AdrienClairembault: I'm not sure a new value is the solution here (I'll let @btry confirm that) but if we add a new constant it should have the same prefix name `STATUS_XXX`

inc/issue.class.php
```diff
@@ -1147,7 +1147,7 @@ static function getReopenableStatusArray() {
 
    static function getAllStatusArray($withmetaforsearch = false) {
       $ticket_status = Ticket::getAllStatusArray($withmetaforsearch);
-      $form_status = [PluginFormcreatorFormAnswer::STATUS_WAITING, PluginFormcreatorFormAnswer::STATUS_ACCEPTED, PluginFormcreatorFormAnswer::STATUS_REFUSED];
+      $form_status = [PluginFormcreatorFormAnswer::STATUS_APPROVAL, PluginFormcreatorFormAnswer::STATUS_ACCEPTED, PluginFormcreatorFormAnswer::STATUS_REFUSED];
```
AdrienClairembault, line 1150: I think `STATUS_APPROVAL` doesn't exist.

inc/issue.class.php
```diff
@@ -1134,7 +1134,7 @@ static function getNewStatusArray() {
    }
 
    static function getPendingStatusArray() {
-      return [Ticket::WAITING, PluginFormcreatorFormAnswer::STATUS_WAITING];
+      return [Ticket::WAITING, PluginFormcreatorFormAnswer::STATUS_APPROVAL];
```
AdrienClairembault, line 1137: I think STATUS_APPROVAL doesn't exist.

AdrienClairembault (changes requested):

inc/formanswer.class.php
```diff
@@ -57,6 +57,7 @@ class PluginFormcreatorFormAnswer extends CommonDBTM
 
    // Values choosen to not conflict with status of ticket constants
    // @see PluginFormcreatorIssue::getNewStatusArray
+   const STATUS_APPROVAL = 101;
    const STATUS_WAITING = 101;
```
Rom1-B, line 61: 101 is used twice

Rom1-B (changes requested):

PR #3479, opened 2024-03-22
fix translation

Rom1-B:
### Changes description

"Access rights" was never translated, even though a translation exists.

image

### Checklist

### References

ref !32199

Closes #N/A


btry:
Indeed, this string does not exists or does no longer exists in GLPI.

PR #3306, opened 2023-06-21
fix: resize dashboard to match GLPI's core

AdrienClairembault:
### Changes description

Resize widgets of the `plugin_formcreator_issue_counters` dashboard to match this sizes used by GLPI's core.

Before:

image

After;

image

### References

!28167

PR #3490, opened 2024-04-09
fix: adjust dashboard widget height

ccailly:
### Checklist

### References

Closes #N/A

Correct dashboard widget height in `install.php`.
Previous pull request, #3306 refactored UI of widgets but missed updating install.php.
A subsequent commit updated `install.php` but set widget height to 1 instead of the correct value 2.

This PR fixes the widget height value to 2.

### DIFF
--- a/install/upgrade_to_2.13.10.php
+++ b/install/upgrade_to_2.13.10.php
@@ -38,16 +38,14 @@ class PluginFormcreatorUpgradeTo2_13_10
    /** @var Migration */
    protected $migration;
 
-   public function isResyncIssuesRequired()
-   {
+   public function isResyncIssuesRequired() {
       return false;
    }
 
    /**
     * @param Migration $migration
     */
-   public function upgrade(Migration $migration)
-   {
+   public function upgrade(Migration $migration) {
       $this->migration = $migration;
       $this->resizeWidgets();
    }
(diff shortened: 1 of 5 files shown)


install/upgrade_to_2.13.10.php
```diff
@@ -0,0 +1,79 @@
+<?php
```
Rom1-B, line 1: It seems simpler and less risky not to resize at all at the update, neither in this version nor in 2.13.7.

Rom1-B (changes requested):

btry (changes requested):
To upgrade to version 2.13.0 the version constant in setup.php must be up to date. Also in install.php there is a matrix which modelizes versions sequence It needs a new entry 2.13.9 => 2.13.10.

Also, I don't expect to release a new version as the plugin reached it end of life and we are implementing forms in GLPI core (v11).

Rom1-B (approved):

ccailly:
As discussed with @Rom1-B above and IRL, it's probably wiser not to change the widget size during an update.
It's possible that some users may have manually modified their widgets, so only new installations of the plugin will benefit from the new widget size.

btry:
Hi 

> As discussed with @Rom1-B above and IRL, it's probably wiser not to change the widget size during an update.
It's possible that some users may have manually modified their widgets, so only new installations of the plugin will benefit from the new widget size.

Actually, the plugin takes measures to prevent users to modify its mini dashboard. Noone should be able to modify it (neither super-admins). This is why it should be safe to modify it programmatically

AdrienClairembault (changes requested):
Agreed with @btry.

The `resizeWidgets` function must not be deleted.

AdrienClairembault (approved):

btry (approved):

PR #3560, opened 2025-01-14
Fix FormcreatorIssues requester not updated

RomainLvr:
### Changes description

When the requester of a ticket is changed to a new requester, the original writer continues to see the ticket in the simplified and extended service catalog interface.
When the Simplified/Extended Service Catalog is changed to GLPI Support, the original ticket writer no longer sees the ticket.

This is due to the requester in the glpi_plugin_formcreator_issues table not being updated when the ticket's actors are updated, as the hook on Ticket_User is not triggered.

### Checklist

### References

It fixes !35730

### DIFF
--- a/hook.php
+++ b/hook.php
@@ -443,6 +443,19 @@ function plugin_formcreator_hook_add_ticket(CommonDBTM $item) {
    ]);
 }
 
+function plugin_formcreator_hook_update_ticket_actors(CommonITILActor $item)
+{
+    if ($item::$itemtype_1 == Ticket::class) {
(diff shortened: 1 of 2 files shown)


btry (approved):
LGTM

stonebuzz (approved):

Rom1-B (approved):

RomainLvr:
do I provide the patch to the customer for validation?

btry:
@RomainLvr I usually validate patches with bug reporter then merge once the fix is approved.

PR #3561, opened 2025-01-22
fix(target) Removes the default user when a group requester is defined in targets

MyvTsv:
### Changes description

If a group is set to requester, but no users are specified by default for the target tickets, the created ticket will add the form’s author as requester in addition to the group when it is created.

Capture d’écran du 2025-01-22 10-29-52

Capture d’écran du 2025-01-22 10-30-48

### Checklist

### References

Ticket !36028


Rom1-B (approved):

MyvTsv:

stonebuzz (approved):

btry (approved):

PR #3563, opened 2025-01-29
fix: set the default value of the dropdown as empty string

MyvTsv:
### Changes description

the mandatory dropdown created with the plugin fields was not checked if the value was empty or not. In the database, the value could be a null value, an empty string or 0 and only the case where the value is equal to 0 is checked which validates the other two cases.

### Checklist

### References

#36030

### DIFF
--- a/inc/field/fieldsfield.class.php
+++ b/inc/field/fieldsfield.class.php
@@ -208,13 +208,19 @@ class FieldsField extends PluginFormcreatorAbstractField
          $field['value'] = $value;
       } else {
          //get default value
-         if ($this->getField()->fields['default_value'] !== "") {
+         if (
+            $this->getField()->fields['default_value'] !== ""
+            && $this->getField()->fields['default_value'] !== null
+            && $this->getField()->fields['default_value'] !== 0
+         ) {
             $value = $this->getField()->fields['default_value'];
             // shortcut for date/datetime
             if (in_array($this->getField()->fields['type'], ['date', 'datetime'])
                   && $value == 'now') {
                $value = $_SESSION["glpi_currenttime"];
             }
+         } else {
+            $value = '';
          }
          $field['value'] = $value;
       }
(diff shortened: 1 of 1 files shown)


inc/field/fieldsfield.class.php
```diff
@@ -208,13 +208,19 @@ public function prepareHtmlField($fieldName, $canedit = true, $value = '') {
          $field['value'] = $value;
       } else {
          //get default value
-         if ($this->getField()->fields['default_value'] !== "") {
+         if (
+            $this->getField()->fields['default_value'] !== ""
+            && $this->getField()->fields['default_value'] !== null
+            && $this->getField()->fields['default_value'] !== 0
+         ) {
```
stonebuzz, line 215:
```suggestion
         if (!empty($this->getField()->fields['default_value'])) {
```

can be simplified

inc/field/fieldsfield.class.php
```diff
@@ -506,7 +512,7 @@ public function isValid(): bool {
    private function isAdditionalFieldEmpty(): bool {
       switch ($this->getField()->fields['type']) {
          case 'dropdown':
-            return $this->getField()->fields['mandatory'] && $this->value == 0;
+            return $this->getField()->fields['mandatory'] && ($this->value == 0 || $this->value == '' || $this->value == null);
```
stonebuzz, line 515:
```suggestion
            return $this->getField()->fields['mandatory'] && empty($this->value);
```

stonebuzz (changes requested):

stonebuzz:
@MyvTsv 

Can you adapt it and send it to the customer for validation?

inc/field/fieldsfield.class.php
```diff
@@ -208,11 +208,7 @@ public function prepareHtmlField($fieldName, $canedit = true, $value = '') {
          $field['value'] = $value;
       } else {
          //get default value
-         if (
-            $this->getField()->fields['default_value'] !== ""
-            && $this->getField()->fields['default_value'] !== null
-            && $this->getField()->fields['default_value'] !== 0
-         ) {
+         if (!empty($this->getField()->fields['default_value'])) {
```
btry, line 211: Should the string "0" be considered empty ?

MyvTsv:

Rom1-B (approved):

PR #3565, opened 2025-02-11
Fix: remove unrelated items when a ticket target is set

MyvTsv:
### Changes description

This PR prevents the user from selecting a GLPI itemtype if it is declared as not associable to a ticket and that the form has for purpose to create a ticket.

Form with no ticket target to specify
Capture d’écran du 2025-02-11 11-27-07 

Form with ticket target to specify
Capture d’écran du 2025-02-11 11-27-41

### References

Closes #36317

### DIFF
--- a/inc/field/dropdownfield.class.php
+++ b/inc/field/dropdownfield.class.php
@@ -246,6 +246,7 @@ class DropdownField extends PluginFormcreatorAbstractField
                $canViewAllHardware = Session::haveRight('helpdesk_hardware', pow(2, Ticket::HELPDESK_ALL_HARDWARE));
                $canViewMyHardware = Session::haveRight('helpdesk_hardware', pow(2, Ticket::HELPDESK_MY_HARDWARE));
                $canViewGroupHardware = Session::haveRight('show_group_hardware', '1');
+               $target_ticket = (new \PluginFormcreatorTargetTicket())->getTargetsForForm($form->getID());
                $groups = [];
                if ($canViewGroupHardware) {
                   $groups = $this->getMyGroups(Session::getLoginUserID());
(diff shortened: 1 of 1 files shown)


inc/field/dropdownfield.class.php
```diff
@@ -266,7 +267,7 @@ public function buildParams($rand = null) {
                   ];
                }
                // Check if helpdesk availability is fine tunable on a per item basis
-               if (Session::getCurrentInterface() == "helpdesk" && $DB->fieldExists($itemtype::getTable(), 'is_helpdesk_visible')) {
+               if ($DB->fieldExists($itemtype::getTable(), 'is_helpdesk_visible') && count($target_ticket) > 0) {
```
Rom1-B, line 270: is the count() necessary? wouldn't that be enough?:
```suggestion
               if ($DB->fieldExists($itemtype::getTable(), 'is_helpdesk_visible')) {
```

Rom1-B (changes requested):

inc/field/dropdownfield.class.php
MyvTsv: This allows you to check whether the form is intended to create a ticket or not. If a form is not intended to create a ticket, then I take into account the elements that are not associable to a ticket.

Rom1-B (approved):

stonebuzz (approved):

stonebuzz:
From 'tbugier'

_I'm not sure that this pr is justified: a form can generate tickets, or not, or even more twisted, depending on the data entered by the requester.
Checking that there's a target ticket isn't enough, you'd also have to check that the ticket will be generated. And this can change when the questions are modified by the requester. _

stonebuzz:
@btry 

After internal discussions, the main issue identified is the visibility of certain elements during the form preview as an administrator, even though these elements cannot be associated with a ticket. Administrators can see these elements, unlike self-service users, due to the condition `Session::getCurrentInterface() == "helpdesk"`.  

Ensuring that the ticket preview matches the form view seen by self-service users is not an issue and allows for consistency between both interfaces. Therefore, we will adjust the PR to remove this verification and ensure a uniform display.

In any case, regardless of the target (change, problem, or ticket), the self-service currently does not have access to elements that cannot be associated with a ticket. This behavior will remain unchanged after this PR.

Rom1-B (approved):

stonebuzz:
@MyvTsv can you propose the related patch to customer ?

MyvTsv:

btry:
Hi

After a test, is_helpdesk_visible state applies to standard interface. this patch is ok for me.

PR #3572, opened 2025-02-27
fix: display supplier input field when selecting a supplier from the …

MyvTsv:
### Changes description

It's fixes !36575

This RP corrects the display of the supplier drop-down menu for observer actors, so that it is no longer visible when creating an observer target in a form because suppliers cannot be observers in a ticket.

Capture d’écran du 2025-03-04 09-27-31

### Checklist

### DIFF
--- a/inc/abstractitiltarget.class.php
+++ b/inc/abstractitiltarget.class.php
@@ -1807,6 +1807,8 @@ SCRIPT;
             break;
          case CommonITILActor::OBSERVER:
             $type = 'watcher';
+            unset($dropdownItems[PluginFormcreatorTarget_Actor::ACTOR_TYPE_SUPPLIER]);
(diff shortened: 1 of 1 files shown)


inc/abstractitiltarget.class.php
```diff
@@ -1999,7 +1999,7 @@ protected function showActorSettingsForType($actorType, array $actors) {
       );
       echo '</div>';
 
-      if ($actorType == CommonITILActor::ASSIGN) {
+      if ($actorType == CommonITILActor::ASSIGN || $actorType == CommonITILActor::OBSERVER) {
```
Rom1-B, line 2002: Can a supplier be an observer on a ticket?

Natively, it seems that they cannot, just as they cannot be a requester. They can only be assigned.

Rom1-B (changes requested):

Rom1-B (approved):

MyvTsv:
Patch works for client

stonebuzz (approved):

PR #3575, opened 2025-03-07
fix: backslashes in translations not working in textarea

MyvTsv:
### Changes description

This PR corrects the translation problem if the text of a textarea to translate contains backslashes

Original :
Capture d’écran du 2025-03-07 10-35-19

Translation :
Capture d’écran du 2025-03-07 10-34-34

### Checklist

Closes !36722

### DIFF
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -402,7 +402,9 @@ PluginFormcreatorTranslatableInterface
     * @return array the modified $input array
     */
    public function prepareInputForAdd($input) {
-      $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
+      if (!empty($input['default_values'])) {
+         $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
+      }
 
       if (!$this->skipChecks) {
          $input = $this->checkBeforeSave($input);
@@ -454,7 +456,9 @@ PluginFormcreatorTranslatableInterface
    public function prepareInputForUpdate($input) {
       // global $DB;
 
-      $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
+      if (!empty($input['default_values'])) {
+         $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
+      }
       if (!$this->skipChecks) {
          if (!isset($input['plugin_formcreator_sections_id'])) {
             $input['plugin_formcreator_sections_id'] = $this->fields['plugin_formcreator_sections_id'];
--- a/inc/question.class.php
+++ b/inc/question.class.php
@@ -402,7 +402,6 @@ PluginFormcreatorTranslatableInterface
     * @return array the modified $input array
     */
    public function prepareInputForAdd($input) {
-
       if (!$this->skipChecks) {
          $input = $this->checkBeforeSave($input);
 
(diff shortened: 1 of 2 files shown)


stonebuzz (approved):
seems good for me (please note that this problem only occurs on textarea)

inc/question.class.php
```diff
@@ -402,6 +402,8 @@ private function checkBeforeSave($input) : array {
     * @return array the modified $input array
     */
    public function prepareInputForAdd($input) {
+      $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
```
Rom1-B, line 405:
```suggestion
      if (!empty($input['default_values'])) {
         $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
      }
```

inc/question.class.php
```diff
@@ -452,6 +454,7 @@ public function prepareInputForAdd($input) {
    public function prepareInputForUpdate($input) {
       // global $DB;
 
+      $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
```
Rom1-B, line 457:
```suggestion
      if (!empty($input['default_values'])) {
         $input['default_values'] = str_replace('\r\n', '', $input['default_values']);
      }
```

Rom1-B (changes requested):

btry (commented):
This change might break setting default values for multiselect questions. To secure the patch i would recommend to check the question type before calling str_replace.

inc/question.class.php
```diff
@@ -456,7 +456,7 @@ public function prepareInputForAdd($input) {
    public function prepareInputForUpdate($input) {
       // global $DB;
 
-      if (!in_array($input['fieldtype'], ['checkboxes', 'multiselect'])) {
+      if (isset($input['fieldtype']) && isset($input['default_values']) && !in_array($input['fieldtype'], ['checkboxes', 'multiselect'])) {
```
btry, line 459: here, if ```$input['fieldtype']``` is not set, you have to read ```$this->fields['fieldtype']```. If not, you may do your str_replace on the types you want to exclude.

Also if @stonebuzz is right, it would be safer to check if the type is ```textarea``` instead of checking it is not one of some types.

inc/question.class.php
```diff
@@ -402,6 +402,10 @@ private function checkBeforeSave($input) : array {
     * @return array the modified $input array
     */
    public function prepareInputForAdd($input) {
+      if (isset($input['fieldtype']) && isset($input['default_values']) && !in_array($input['fieldtype'], ['checkboxes', 'multiselect'])) {
```
Rom1-B, line 405: Does the initial issue only concern textarea? So it should only be applied in this case and no others.

Something like this (not tested):
```suggestion
      if (
         isset($input['fieldtype'])
         && $input['fieldtype'] == 'textarea'
         && !empty($input['default_values'])
      ) {
```

`empty()` avoids trying to `str_replace` if the key doesn't exist or its value is empty

Rom1-B (changes requested):

Rom1-B (approved):

inc/question.class.php
```diff
@@ -402,9 +402,14 @@ private function checkBeforeSave($input) : array {
     * @return array the modified $input array
     */
    public function prepareInputForAdd($input) {
+
```
stonebuzz, line 405: useless

stonebuzz (changes requested):

stonebuzz (approved):

btry (approved):

PR #3578, opened 2025-03-12
Fix - Save button back in form validation

RomainLvr:
### Checklist

Restore save button disappeared from fix : https://github.com/glpi-project/glpi/pull/18286

Capture d’écran du 2025-03-12 10-38-58


RomainLvr:
> Hi
> 
> Why is it needed to show the save button here ? The user must accept / refuse or edit the form answer here.

The client uses this button to add comments for other validators. A kind of note to let other validators know that the form is being processed, or that sort of thing.

PR #3618, opened 2025-10-20
fix(import): Export sections in logical order to fix condition import

MyvTsv:
### Changes description

It fixes !39806

During import, if a question in section A contains a condition that references a question in section B, but section B appears after section A in the JSON array (due to the order of the IDs), the UUID linker cannot resolve the reference. The order of the sections in the array is based on the ID rather than the order of the sections in the form.

image

### DIFF
--- a/inc/exportabletrait.class.php
+++ b/inc/exportabletrait.class.php
@@ -55,9 +55,17 @@ trait PluginFormcreatorExportableTrait
          }
          $export[$key] = [];
          foreach ($itemtypes as $itemtype) {
-            $allSubItems = $DB->request($itemtype::getSQLCriteriaToSearchForItem($this->getType(), $this->getID()));
-            $list = [];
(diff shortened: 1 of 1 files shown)


inc/exportabletrait.class.php
```diff
@@ -55,9 +55,17 @@ public function exportChildrenObjects(array $subItems, array $export, bool $remo
...
-            $list = [];
+            $criteria = $itemtype::getSQLCriteriaToSearchForItem($this->getType(), $this->getID());
+
+            // Add ORDER BY clause if the itemtype has an 'order' field
+            // This ensures sections and other ordered items are exported in the correct order
             $subItem = new $itemtype();
+            if ($subItem->isField('order')) {
+               $criteria['ORDER'] = ['order ASC'];
+            }
+
```
Rom1-B:
```suggestion

```

stonebuzz (approved):
seems good for me

Rom1-B (approved):

PR #3621, opened 2025-11-05
Fix (Uninstall) - Restore mini-dashboard delete function

RomainLvr:
When uninstalling the plugin in v3.0.0, the mini-dashboard entry created by the plugin is not deleted, which causes problems with the mini-dashboard not working after uninstalling the plugin.

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,8 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<phpunit bootstrap="tests/bootstrap.php" colors="true" testdox="true">
-   <testsuites>
-      <testsuite name="Units">
-         <directory>tests/units</directory>
-      </testsuite>
-   </testsuites>
-</phpunit>


src/Install.php
```diff
@@ -491,6 +491,27 @@ protected function deleteTables() {
...
+
+      $dashboard->delete([
+         'key' => 'plugin_formcreator_issue_counters'
+      ]);
+      if ($dashboard->getFromDB('plugin_formcreator_issue_counters')) {
+         // Failed to delete the dashboard
+         $res = false;
+      }
+
+      return $res;
```
Rom1-B:
```suggestion
      return $dashboard->deleteByCriteria([
         'key' => 'plugin_formcreator_issue_counters'
      ]);
```

Rom1-B (changes requested):

Rom1-B (approved):

PR #3642, opened 2026-04-09
Fix: upgrade function never called

Rom1-B:
### Changes description

The table schema update was no longer being called.

### Checklist

### References

Closes #N/A


hook.php
```diff
@@ -49,11 +49,6 @@ function plugin_formcreator_install() {
    // Use the unified Install class
    $install = new Install();
 
-   if (!$install->install($migration)) {
-      $migration->displayMessage("ERROR: Formcreator installation failed");
-      return false;
-   }
```
btry, line 55: I would suggest to forbid installation, as you want to allow upgrade only.

the code below blocks installation if there are tables of the plugin in the database (see implementation for details).

```suggestion
   if (!$install->isPluginInstalled() {
      $migration->displayMessage("ERROR: Installation not allowed");
      return false;
   }
```

stonebuzz (approved):

btry (approved):

PR #3647, opened 2026-06-26
Fix - Prevent TypeError in CheckboxesField::regex() when stored value is invalid JSON

RomainLvr:
### Checklist

## Description

- It fixes !44948
- Here is a brief description of what this PR does

When rendering the `plugin_formcreator_solved_issues` dashboard widget, a `TypeError` is thrown:

```php
preg_grep(): Argument #2 ($array) must be of type array, null given
```

This happens when a form answer contains a checkbox field with a regex visibility condition, and the stored JSON value is corrupted or invalid. `json_decode()` returns `null` on invalid input, leaving `$this->value = null`, which then causes `preg_grep()` to throw.

A secondary bug was also found in `LdapselectField::regex()`, which incorrectly used `preg_grep()` on a plain string value instead of `preg_match()`.

### Changes

- `CheckboxesField::deserializeValue()`: add `is_array()` guard after `json_decode()` so `$this->value` always ends up as an array, even when the stored data is malformed. This mirrors the existing pattern in `ActorField`.
- `LdapselectField::regex()`: replace `preg_grep()` with `preg_match()`, consistent with `RadiosField` which holds a single string value.
- `tests/CheckboxesField.php`: add `testRegex()` covering valid match, no match, invalid JSON, and JSON-null cases.

### DIFF
--- a/inc/field/checkboxesfield.class.php
+++ b/inc/field/checkboxesfield.class.php
@@ -141,9 +141,10 @@ class CheckboxesField extends PluginFormcreatorAbstractField
    }
 
    public function deserializeValue($value) {
-      $this->value = ($value !== null && $value !== '')
+      $decoded = ($value !== null && $value !== '')
          ? json_decode($value)
          : [];
+      $this->value = is_array($decoded) ? $decoded : [];
    }
 
    public function getValueForDesign(): string {
--- a/inc/field/ldapselectfield.class.php
+++ b/inc/field/ldapselectfield.class.php
@@ -221,7 +221,7 @@ class LdapselectField extends SelectField
    }
 
    public function regex($value): bool {
-      return (preg_grep($value, $this->value)) ? true : false;
+      return preg_match($value, $this->value) ? true : false;
    }
 
    public function isPublicFormCompatible(): bool {
--- a/tests/3-unit/GlpiPlugin/Formcreator/Field/CheckboxesField.php
+++ b/tests/3-unit/GlpiPlugin/Formcreator/Field/CheckboxesField.php
@@ -602,6 +602,47 @@ class CheckboxesField extends CommonAbstractFieldTestCase {
       $this->boolean($output)->isTrue();
    }
 
+   public function providerRegex() {
+      yield 'valid json value matches regex' => [
+         'value'    => '["foo","bar"]',
+         'pattern'  => '/foo/',
+         'expected' => true,
+      ];
+      yield 'valid json value does not match regex' => [
+         'value'    => '["foo","bar"]',
+         'pattern'  => '/baz/',
+         'expected' => false,
+      ];
+      yield 'corrupted json value does not throw and returns false' => [
+         'value'    => 'not_valid_json',
+         'pattern'  => '/foo/',
+         'expected' => false,
+      ];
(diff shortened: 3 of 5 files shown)


stonebuzz (commented):
LGTM

inc/field/ldapselectfield.class.php
```diff
@@ -221,7 +221,7 @@ public function lessThan($value): bool {
    }
 
    public function regex($value): bool {
-      return (preg_grep($value, $this->value)) ? true : false;
+      return preg_match($value, $this->value) ? true : false;
```
Rom1-B, line 224: `regex()` behavior changed (`preg_grep` → `preg_match`), but `tests/3-unit/GlpiPlugin/Formcreator/Field/LdapSelectField.php` has no test for `regex()` at all, before or after this fix. Can you add a case mirroring `providerRegex` in `CheckboxesField.php`?

inc/field/checkboxesfield.class.php
```diff
@@ -141,9 +141,10 @@ public function serializeValue(PluginFormcreatorFormAnswer $formanswer): string
    }
 
    public function deserializeValue($value) {
-      $this->value = ($value !== null && $value !== '')
+      $decoded = ($value !== null && $value !== '')
          ? json_decode($value)
          : [];
+      $this->value = is_array($decoded) ? $decoded : [];
```
Rom1-B, line 147: `providerRegex` covers `null`, `''`, invalid JSON string, and the JSON `null` literal, but not other valid-JSON-non-array values (e.g. `'"a string"'`, `'42'`, `'{"a":1}'`) that also hit the `is_array($decoded) ? $decoded : []` branch. Not required given the generic array check, but worth one more case for full branch coverage.

Rom1-B (changes requested):

tests/3-unit/GlpiPlugin/Formcreator/Field/CheckboxesField.php
```diff
@@ -602,6 +602,47 @@ public function testCanRequire() {
...
+      yield 'corrupted json value does not throw and returns false' => [
+         'value'    => 'not_valid_json',
+         'pattern'  => '/foo/',
+         'expected' => false,
+      ];
+      yield 'null-decoded json value does not throw and returns false' => [
+         'value'    => 'null',
+         'pattern'  => '/foo/',
+         'expected' => false,
+      ];
```
stonebuzz, line 625: `providerRegex` doesn't cover valid-JSON non-array values (e.g. `'"a string"'`, `'42'`, `'{"a":1}'`) that also reach the `is_array($decoded) ? $decoded : []` fallback branch.

```suggestion
      yield 'null-decoded json value does not throw and returns false' => [
         'value'    => 'null',
         'pattern'  => '/foo/',
         'expected' => false,
      ];
      yield 'valid json non-array value does not throw and returns false' => [
         'value'    => '"a string"',
         'pattern'  => '/foo/',
         'expected' => false,
      ];
```

stonebuzz (changes requested):

Rom1-B (approved):
Sounds good to me. Can you ask the client to approve it?

The test are failing. Can you fix that?

stonebuzz (approved):

